PR #7, opened 2025-07-24
getBoardingTxs: increase poll ticker interval

louisinger:
@altafan please review

## Summary by CodeRabbit

* **Chores**
  * Reduced the frequency of polling for boarding transactions, resulting in fewer updates per minute.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1117,7 +1117,7 @@ func (a *arkClient) refreshVtxoDb(spendableVtxos, spentVtxos []types.Vtxo) error
 }
 
 func (a *arkClient) listenForBoardingTxs(ctx context.Context) {
-	ticker := time.NewTicker(2 * time.Second)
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -1117,7 +1117,7 @@ func (a *arkClient) refreshVtxoDb(spendableVtxos, spentVtxos []types.Vtxo) error
 }
 
 func (a *arkClient) listenForBoardingTxs(ctx context.Context) {
-	ticker := time.NewTicker(2 * time.Second)
+	ticker := time.NewTicker(time.Minute)
```
altafan, line 1120: if we do a call every 10s, we would do 6 calls per minutes, 360 calls per hour, 8640 calls per day, 259200 calls per month. The limit is 500k per month so we should be ok wihtout making the detection of boarding utxo too much slow.

PR #10, opened 2025-08-01
Implement gRPC reconnection

tiero:
## Summary
- add connection state monitoring and reconnection for the Ark gRPC clients
- dial using stored options and ensure the connection is ready before each RPC

## Testing
- `go test ./...`

------
https://chatgpt.com/codex/tasks/task_b_688d3187ee80832c8526d229af5a7195

### DIFF
--- a/client/grpc/client.go
+++ b/client/grpc/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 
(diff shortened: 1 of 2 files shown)


altafan (changes requested):
With the following changes the reconnection would be done if and only if the user tries to call an endpoint.
There may be existing streams that have been closed and never re-opened.

The logic must run in the background as soon as the connection drops, and possibly we should also restore all streams to make the user feel really unaware of what's happening under the hood.

PR #70, opened 2025-12-11
Add reconnection logic to ws tracking

altafan:
This adds reconnection logic in case the connection drops, and ensures all address subscriptions are restored as well.

This is an enhancement of and replaces #67, as instead of just returning an error in case of a connection drop, we try to restore the connection until the user stops the execution of the explorer service.

Please @sekulicd review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket reliability with exponential backoff retries and clearer retry/abort behavior on cancellation.
  * Better differentiation between timeout vs. permanent close errors to reduce noisy broadcasts and enable targeted recovery.

* **Improvements**
  * More robust subscribe/unsubscribe with explicit subscription confirmation and explicit unsubscription handling.
  * Centralized reconnection/resubscription flows, reduced per-address concurrency, and clearer connection availability management.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/explorer/mempool/connection_pool.go
+++ b/explorer/mempool/connection_pool.go
@@ -139,7 +139,7 @@ func (cp *connectionPool) pushAddress(address string) (int, error) {
 	defer cp.mu.Unlock()
 
 	if len(cp.connections) == 0 {
-		return -1, fmt.Errorf("no connections avaialble")
+		return -1, fmt.Errorf("no connections available")
 	}
 
 	// Select the first available connection without an address assigned
@@ -149,7 +149,7 @@ func (cp *connectionPool) pushAddress(address string) (int, error) {
 	})
 	// If connections are all taken, reject the request
 	if idx < 0 {
-		return -1, fmt.Errorf("no more connections availble")
+		return -1, fmt.Errorf("no more connections available")
 	}
 
 	connId := conns[idx].id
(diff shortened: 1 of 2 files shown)


explorer/mempool/connection_pool.go
```diff
@@ -91,7 +119,8 @@ func (cp *connectionPool) addConnection() error {
 	return nil
 }
 
-func (cp *connectionPool) resetConnection(wsConn *websocketConnection) {
+// resetConnection closes and removes the given connection from the pool, and adds a new one
+func (cp *connectionPool) resetConnection(wsConn *websocketConnection) error {
```
sekulicd: I think we have race condition in here since we the lock is released between deletion and recreation

  ```
func (cp *connectionPool) resetConnection(wsConn *websocketConnection) error {
      cp.mu.Lock()
      wsConn.conn.Close()
      delete(cp.connections, wsConn.id)
      delete(cp.connectionsByAddress, wsConn.address.get())
      cp.noMoreConnections = false
      cp.mu.Unlock()  // ← LOCK RELEASED HERE

      return cp.addConnection()  // ← NEW lock acquired inside
  }
```

altafan: mmm i don't think so? we don't really care which conn arrives to addConnection when it is reset, it is just enough one is handled,  the others are going to be queued and solved once the connectivity issues are solved. Re: https://github.com/arkade-os/go-sdk/pull/70#discussion_r2613733045

PR #88, opened 2026-01-30
improved asset fee calculation

aruokhai:


louisinger:
overlaping with revert in https://github.com/arkade-os/go-sdk/pull/58

PR #107, opened 2026-03-09
fix: remove global log mutation and fix example race conditions

### LINKED ISSUE #106: Bug: SDK overwrites application log level and example scripts fail from race conditions
### 1. Global Logger Mutation
**The Problem:** Calling `Unlock()` in `base_client.go` explicitly sets the global `logrus` level to `DebugLevel` or `ErrorLevel`.
**The Impact:** The SDK forcefully overwrites the logging configuration of any parent application importing it. Libraries should not mutate global state.

### 2. Race Conditions in Examples
**The Problem:** Example scripts (like `alice_to_bob.go`) do not wait for the background wallet sync to finish after calling `Unlock()`. 
**The Impact:** The scripts race ahead and try to spend funds before the internal database is populated.

### How to Reproduce:
For this to work, #105 should have been merged to solve the panic bug. Once that is done, you can proceed with:
1. Run `example/alice_to_bob/alice_to_bob.go`
2. Logs  after the `Unlock()`  for alice won't show 
3. It exits with error "wallet is still syncing"

Image

Dunsin-cyber:
fixes #106 

## Changes
- **`base_client.go`**: Removed logic that forcefully set `logrus` levels.
- **`example/alice_to_bob/alice_to_bob.go`**: Added await logic using `<-client.IsSynced(ctx)` for both Alice and Bob clients to ensure the local database is populated before proceeding.

## Before
alice-bob-err-2

## Now
working example

## How to Test
1. Run `go run example/alice_to_bob/alice_to_bob.go ` from the root dir

## Summary by CodeRabbit

* **Bug Fixes**
  * Added wallet synchronization checks after unlock with timeouts, error handling, and success confirmation to ensure subsequent actions run reliably.

* **Changes**
  * Made log-level configuration updates during unlock concurrency-safe to prevent race conditions affecting runtime logging.

### DIFF
--- a/base_client.go
+++ b/base_client.go
@@ -39,6 +39,7 @@ const (
 var (
 	ErrAlreadyInitialized = fmt.Errorf("client already initialized")
 	ErrNotInitialized     = fmt.Errorf("client not initialized")
+	logMu sync.Mutex
 )
 
 type ClientOption func(*arkClient)
@@ -110,6 +111,13 @@ func (a *arkClient) Unlock(ctx context.Context, password string) error {
 	if _, err := a.wallet.Unlock(ctx, password); err != nil {
 		return err
 	}
+
+	logMu.Lock()
+	log.SetLevel(log.DebugLevel)
+	if !a.verbose {
+		log.SetLevel(log.ErrorLevel)
+	}
+	logMu.Unlock()
 
 	a.dbMu = &sync.Mutex{}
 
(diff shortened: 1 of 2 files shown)


altafan:
@Dunsin-cyber we included these changes in #103. We've almost finalized an internal refactor and it was easier and faster to include the same changes there as part of the pr review 🙏. Thank you again but i'm closing this

PR #143, opened 2026-04-17
HD Wallet

sekulicd:
## Summary

This PR adds SDK-managed HD wallet support and wires it through the full client lifecycle.

This introduces a new `wallet/hdwallet` implementation, integrates it into `Init` / `LoadArkClient` / `Unlock`, adds HD-specific restore and discovery logic, and updates the e2e suite to run against HD wallets only.

## What’s new

- add BIP32/BIP39 hierarchical deterministic wallet support in `wallet/hdwallet`
- derive all Ark address families from one key per index using:
  - `m/83696968'/11811'/0/{index}`
- keep derivation aligned with the Rust SDK convention
- persist the encrypted HD wallet state in the existing config-store-backed data dir
- support:
  - mnemonic creation/restoration
  - raw seed restoration
  - encrypted master key + mnemonic persistence
- add `WithHDWallet(store)` init option
- add `WithGapLimit(n)` client option for startup key discovery
- bump `client-lib` / `ark-lib` / `api-spec` to `altafan/ark@86c10c7f3076`

## SDK integration

- `Init` can now construct and inject the built-in HD wallet when `WithHDWallet(...)` is used
- `LoadArkClient` reconstructs the HD wallet from persisted wallet type/config and loads it through `LoadArkClientWithWallet(...)`
- `Unlock` performs HD discovery before first sync and avoids pending-tx finalization / VTXO refresh paths when the wallet has no keys yet
- `GetAddresses`, `NewOffchainAddress`, `NewBoardingAddress`, and `NewOnchainAddress` now work with HD-backed fresh key allocation

## Discovery and recovery

Startup discovery uses a gap-limit scan and expands the known key range from both Ark and Bitcoin-visible activity:

- offchain VTXOs via the Ark indexer
- boarding UTXOs via explorer lookups
- redemption/materialized VTXO outputs via explorer lookups
- direct onchain UTXOs via explorer lookups

This lets restored HD wallets recover used keys and rebuild mixed offchain/onchain state before the client reports itself synced.

## Onchain tracking and signing

- add dynamic onchain address tracking so newly derived HD addresses are subscribed after startup as they are allocated
- restore tracked onchain UTXOs from full address transaction history instead of only current UTXO snapshots
- preserve richer explorer-derived UTXO details during rebuild, including spent state and raw tx data
- implement HD signing for:
  - Ark tapscript spends
  - direct wallet-owned Taproot key-path spends

## Tests

- add dedicated HD wallet unit coverage
- add HD e2e coverage for:
  - fresh address allocation
  - gap-limit recovery
  - mixed offchain/onchain restore
  - HD event streams
- migrate the e2e suite to HD-backed clients
- simplify e2e coverage to SQL store only
- set `make integrationtest` to run with a 20 minute `go test` timeout

## Release Notes

* **New Features**
  * Added hierarchical deterministic (HD) wallet support with BIP39 mnemonic generation, encryption, and backup/restore functionality.
  * Added automatic address discovery and key management for HD wallets with configurable gap limits.
  * Added new transaction methods: `RegisterIntent`, `DeleteIntent`, `CompleteUnroll`, `WithdrawFromAllExpiredBoardings`, and `OnboardAgainAllExpiredBoardings`.

* **Documentation**
  * Updated README with comprehensive HD wallet usage guide, including initialization, backup, and restoration examples.

* **Tests**
  * Added extensive HD wallet test coverage across address allocation, restoration, and mixed fund recovery scenarios.
  * Refactored end-to-end tests for simplified and more deterministic execution.


go.mod
```diff
@@ -1,9 +1,17 @@
...
 
 replace github.com/btcsuite/btcd/btcec/v2 => github.com/btcsuite/btcd/btcec/v2 v2.3.3
 
+replace github.com/arkade-os/arkd/api-spec => github.com/altafan/ark/api-spec v0.0.0-20260411092811-86c10c7f3076
+
+replace github.com/arkade-os/arkd/pkg/ark-lib => github.com/altafan/ark/pkg/ark-lib v0.0.0-20260411092811-86c10c7f3076
+
+replace github.com/arkade-os/arkd/pkg/client-lib => github.com/altafan/ark/pkg/client-lib v0.0.0-20260411092811-86c10c7f3076
+
+replace github.com/arkade-os/arkd/pkg/errors => github.com/altafan/ark/pkg/errors v0.0.0-20260411092811-86c10c7f3076
```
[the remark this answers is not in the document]
altafan: @sekulicd no need for these replacements. I created a clone of that branch in arkd repo, so you can just drop'em and update the client-lib dep with `go get github.com/arkade-os/arkd/pkg/client-lib@86c10c7f30766f8353ca20b23046f6f29ba90e8a` + `go mod tidy`

wallet/hdwallet/key_service.go
```diff
@@ -0,0 +1,284 @@
...
+// at the given child index.
+func (p *KeyService) DefaultKeyPath(index uint32) string {
+	return fmt.Sprintf("%s/%d", defaultKeyPathPrefix, index)
+}
+
+// NewHDKeyProvider creates a new key provider from a BIP32 master extended key.
+func NewHDKeyProvider(masterKey *hdkeychain.ExtendedKey) *KeyService {
+	keyBasePath, err := parseKeyPathPrefix(defaultKeyPathPrefix)
+	if err != nil {
+		panic(fmt.Sprintf("invalid default HD key path prefix %q: %v", defaultKeyPathPrefix, err))
```
altafan, line 59: no panic please, return error instead

altafan: Also, it seems awkward to check a hardcoded path.

Since the function is exported and I assume it's meant to be reused on purpose (make it private otherwise), then it's prob better to make the path configurable, so users are free to use the path they want, and move the hardcoded one to the client (but we could add a WithWalletDerivationPath option to have it configurable at sdk level too)?

altafan (changes requested):
@sekulicd the logic to retrieve the private keys to be used for signing seems not super efficient (loading all derived pubkeys + retrieving the priv key for all of them -> O(N^2)).

Anyway, considering that we miss the contract manager that is like the "glue" between an address/script and a key index, it may sound counterintuitive, but here i think we have to change our approach:

At the moment, `SignTransaction` expects just a tx to be signed as the arg, meaning the wallet is responsible for identifying which input scripts are owned and retrieving the correct keys to sign them.

We should move this logic out of this component and, therefore, change the definition of `SignTransaction` so that it accepts a map _input index -> key id_ (derivation index, uint32, should probably be enough).  It would live in the client until we have the contract manager, so we could pack in a way that would be easy to drop off later (soon).

This way:
* this component is really responsible just for deriving keys and signing, no input recognition logic or whatever
* this component is agnostic from network and arkade operator. It's just a "bag of keys" as desired
* retrieving the key to sign an input costs O(1) (here, but it will be O(1) also in cotract manager, and therefore O(1) overall)

wallet/hdwallet/key_service.go
sekulicd: Regarding hardcoding path, we have this path in Rust SDK, IMO sdk’s should have aligned derivation path so all sdks could be used to recover.
Wrote this in PR desc:
```
- derive all Ark address families from one key per index using:
m/83696968'/11811'/0/{index}
- keep derivation aligned with the Rust SDK convention
```
@altafan

test/e2e/hd_wallet_test.go
```diff
@@ -0,0 +1,561 @@
...
+	require.NoError(t, err)
+
+	require.Eventually(t, func() bool {
+		bobBalanceAfter, err := bobClientHD.Balance(ctx)
+		if err != nil {
+			return false
+		}
+
+		return bobBalanceAfter.OffchainBalance.Total == bobBalanceBefore.OffchainBalance.Total+20_000
+	}, 30*time.Second, 500*time.Millisecond)
```
altafan, line 135: let's lower the timeout to 10 seconds and increasy the frequency of the poll
```suggestion
	}, 10*time.Second, 200*time.Millisecond)
```

test/e2e/utils_test.go
```diff
@@ -31,102 +39,53 @@ const (
 	explorerUrl = "http://127.0.0.1:3000"
 )
 
-type testStoreBackend struct {
-	name string
+type testHDWallet struct {
+	sdk.ArkClient
+	WalletSvc    wallet.WalletService
+	TransportSvc transport.TransportClient
```
altafan: no need for these params? they can be retrieved with like for example:
```
alice := setupHdWallet(t, "")
wallet := alice.Wallet()
transport := alice.Transport()
```
?

test/e2e/utils_test.go
```diff
@@ -182,6 +154,117 @@ func faucetOffchain(t *testing.T, client sdk.ArkClient, amount float64) clientTy
 	return vtxo
 }
 
+func deriveWalletAddresses(
```
altafan, line 95: why do we need this?

altafan: it's used only in `TestOffchainTx` of `test/e2e/transaction_test.go` to get all derived offchain addresses. Can't we simply use `alice.GetAddresses` instead and drop this function?

wallet/hdwallet/wallet.go
```diff
@@ -0,0 +1,1198 @@
...
+	Type = "hd"
+	// DefaultGapLimit is the number of unused addresses to check during discovery.
+	DefaultGapLimit      = uint32(20)
+	hdWalletStateVersion = uint32(2)
+	passwordSaltSize     = 32
+	pbkdf2Iterations     = 600_000
+)
+
+// Compile-time interface check.
+var _ wallet.WalletService = (*Service)(nil)
```
altafan: drop, we don't need this as we are gonna feed the client-lib with this wallet. That's the smae check at compile time. This is redundant (and also ugly IMO)

wallet/hdwallet/wallet.go
```diff
@@ -0,0 +1,1198 @@
...
+type Service struct {
+	keyProvider         *KeyService
+	store               Store
+	indexer             indexer.Indexer
+	explorer            explorer.Explorer
+	keyPathPrefix       string
+	arkNetwork          arklib.Network
+	signerPubKey        *btcec.PublicKey
+	boardingExitDelay   arklib.RelativeLocktime
+	unilateralExitDelay arklib.RelativeLocktime
```
altafan: this is a bag of keys, doesn't need any of these config args but the `keyPathPrefix`
```suggestion
	keyPathPrefix       string
```

Maybe also the explorer is needed for signing? like it's needed in the client-lib? but all other args can be def dropped

wallet/hdwallet/wallet.go
```diff
@@ -0,0 +1,1198 @@
...
+	Explorer            explorer.Explorer
+	KeyPathPrefix       string
+	ArkNetwork          arklib.Network
+	SignerPubKey        *btcec.PublicKey
+	BoardingExitDelay   arklib.RelativeLocktime
+	UnilateralExitDelay arklib.RelativeLocktime
+}
+
+// NewService creates a new HD wallet service with all known dependencies.
+func NewService(args Args) (*Service, error) {
```
altafan: in light of the comment above, we can drop this `Args` and just have plain function args:

```suggestion
// NewService creates a new HD wallet service with all known dependencies.
func NewService(
    keyPathPrefix string, store Store, explorer explorer.Explorer,
) (*Service, error) {
```

wallet/hdwallet/store.go
```diff
@@ -0,0 +1,164 @@
...
+	PasswordSalt       []byte `json:"password_salt,omitempty"`
+	EncryptedMnemonic  []byte `json:"encrypted_mnemonic,omitempty"`
+	OffchainNextIndex  uint32 `json:"offchain_next_index"`
+}
+
+// Store is the persistence interface for HD wallet state.
+type Store interface {
+	Save(ctx context.Context, state *State) error
+	Load(ctx context.Context) (*State, error)
+}
```
altafan: since the wallet impl explicitly checks for wallet already inizialized in Create, we need here an extra Clear method because the sdk client offers the Reset feature, where one can drop all stored data, included the wallet one and reinitialize the client using the same datadir/storage without require a shutdown of the service
```suggestion
type Store interface {
	Save(ctx context.Context, state *State) error
	Load(ctx context.Context) (*State, error)
    Clear(ctx context.Context)
}
```

wallet/hdwallet/store.go
```diff
@@ -0,0 +1,164 @@
+package hdwallet
```
altafan: this file contains both interface and impl of the store. Ofc we cannot merge this pr with just the in-memory implememtation, therefore better to create a store/ folder with store.go including only the interface and store/in-memory/store.go, store/file/store.go, and so on for the several implementation we want to support now and in the future, similary to how it's been done in client-lib for the single key store.

wallet/hdwallet/store.go
```diff
@@ -0,0 +1,164 @@
...
+type Store interface {
+	Save(ctx context.Context, state *State) error
+	Load(ctx context.Context) (*State, error)
+}
+
+const hdWalletStateFile = "hd_wallet_state.json"
+
+// configStoreBackend implements Store using the client-lib ConfigStore's
+// data directory for file-based persistence.
+type configStoreBackend struct {
```
altafan: i'm pretty sure this is inherited from the client-lib and i'm gonna solve this in pr 1008. Let me explain: originally, as you know, the wallet component in client oib was responsible for deriving addresses and this is exactly what the pr addresses (those apis has been replaced by other to handle keys). This is the very reason why in client-lib the wallet store takes a configStore, that is where all params for GetInfo are stored to derive addresses. This is simply not needed in wallet anymore and i'm gonna fix this as well in pr 1008.

altafan: please clean this up too

wallet/hdwallet/wallet.go
```diff
@@ -0,0 +1,1198 @@
...
+}
+
+// Create initializes the HD wallet using one of three seed modes:
+// - empty seed: generate and return a new BIP39 mnemonic
+// - valid mnemonic: restore from that mnemonic and return it unchanged
+// - hex string: treat it as raw master-seed bytes and return the same hex seed
+func (w *Service) Create(
+	ctx context.Context, password, seed string,
+) (string, error) {
+	w.mu.Lock()
```
altafan, line 63: first thing first, we must make sure the password is not empty
```suggestion
    if len(password) <= 0 {
        return "", fmt.Errorf("missing password")
    }
    
	w.mu.Lock()
```

wallet/hdwallet/wallet.go
altafan: it seems we also need the network (https://github.com/sekulicd/go-sdk/blob/go-sdk-hd-wallet-sync/wallet/hdwallet/wallet.go#L149-L150), fine if you want to keep Args, but let's drop all unnecessary fields from there

wallet/hdwallet/wallet.go
```diff
@@ -0,0 +1,1198 @@
...
+		if err != nil {
+			return "", fmt.Errorf("failed to encrypt mnemonic: %w", err)
+		}
+	}
+
+	passwordSalt, err := randomBytes(passwordSaltSize)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate password salt: %w", err)
+	}
+	passwordVerifier := derivePasswordVerifier(pwd, passwordSalt)
```
altafan: can't we just sha256 hash the password and store it? less things to store?

wallet/hdwallet/wallet.go
```diff
@@ -0,0 +1,1198 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	restored := state.OffchainNextIndex > 0
+	if restored {
+		if err := w.keyProvider.LoadState(*state); err != nil {
+			return false, fmt.Errorf("failed to restore key state: %w", err)
+		}
+	}
```
altafan: This should be moved at the very end of this function, to really ensure that if w.keyProvider is not nil the wallet is already initialized. Atm this is not really true, because it seems things can still fail after that point. BUT, it seems like the impl of LoadState never returns error. Shall we just make it a void function? this way, we could really move this code at the very bottom and can rely on checking w.keyProvider against nil to know whether the wallet is initialized or not.

altafan: we could get rid also of the `lock bool` field at that point

wallet/hdwallet/wallet.go
```diff
@@ -0,0 +1,1198 @@
...
+}
+
+func (w *Service) GetType() string {
+	return Type
+}
+
+// Create initializes the HD wallet using one of three seed modes:
+// - empty seed: generate and return a new BIP39 mnemonic
+// - valid mnemonic: restore from that mnemonic and return it unchanged
+// - hex string: treat it as raw master-seed bytes and return the same hex seed
```
altafan: i'd just drop this option tbh. Let's make it simple and work only with mnemonics

altafan: means we might have to store the encrypted mnemonic in the store and not the encrypted master key unless there's a way to easily go from masterKey to mnemonic

wallet/hdwallet/options.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	if args.id == "" {
+		return "", fmt.Errorf("key id is required")
+	}
+
+	return args.id, nil
+}
+
+func parseOffchainIndex(keyID string) (uint32, error) {
```
altafan: `parseDerivationIndex` seems a more appropriate name for this function
```suggestion
func parseDerivationIndex(keyID string) (uint32, error) {
```

wallet/hdwallet/wallet.go
```diff
@@ -0,0 +1,1198 @@
...
+	return nil
+}
+
+func (w *Service) Unlock(ctx context.Context, password string) (bool, error) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.keyProvider != nil {
+		return true, nil // already unlocked
+	}
```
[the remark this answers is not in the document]
sekulicd: Done

init.go
```diff
@@ -75,10 +112,25 @@ func (a *arkClient) Init(
...
 
+	hasKeys, err := a.walletHasKeys(ctx)
+	if err != nil {
+		return err
+	}
+	if hasKeys {
+		if _, err := a.FinalizePendingTxs(ctx, nil); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
sekulicd: Done

ghost (changes requested):
## 🔍 Arkana Code Review — HD Wallet (#143)

This is protocol-critical code (key derivation, transaction signing, fund custody). Thorough review follows. **Human sign-off required before merge per protocol-critical policy.**

---

### 🔴 CRITICAL

**C1. `signing_keys.go:60-75` — Positional index correlation for onchain address KeyIDs is a funds-loss vector**

Onchain addresses have no `KeyID`. The code infers them by matching array indices with offchain/boarding/redemption address lists:

```go
for i, addr := range onchainAddrs {
    keyID := ""
    if i < len(offchainAddrs) {
        keyID = offchainAddrs[i].KeyID
    }
```

This assumes `getAllocatedAddresses` returns all four address types in lock-step order and equal length. If any address type is ever filtered, skipped, or reordered (e.g., by an upstream `GetAddresses` change), **the wrong signing key binds to an onchain UTXO**. A transaction signed with the wrong key is either invalid (funds stuck) or, in collaborative exit paths, could produce unexpected behavior.

**Fix:** Onchain addresses should carry their own `KeyID`, or maintain an explicit `outputScript → keyID` mapping derived from the key itself.

**C2. `wallet.go:108` — Unsalted single-SHA-256 password hash negates PBKDF2 protection**

```go
passwordHash := sha256.Sum256(pwd)
```

The encryption key uses PBKDF2 with 600k iterations (good), but the password *verification* hash is a single unsalted SHA-256. An attacker with the wallet file brute-forces the SHA-256 hash at billions/sec on commodity GPUs to recover the password, then decrypts the master key. The verification hash is the weakest link.

**Fix:** Remove `PasswordHash` entirely. Verify passwords by attempting AES-GCM decryption — if authentication fails, the password is wrong. This is already the behavior in `Unlock` at line 200. The `PasswordHash` field is redundant and harmful.

**C3. `wallet.go:573-586` — `persistState` sends zero-value credentials to store, relying on implicit merge contract**

```go
return w.store.Save(ctx, walletstore.State{
    NextIndex: w.keyProvider.GetNextKeyIndex(),
})
```

`EncryptedMasterKey`, `EncryptedMnemonic`, `PasswordHash` are all zero-valued. Current store implementations (file and in-memory) happen to merge correctly, but the `Store.Save` interface contract says "stores the given state if not yet persisted otherwise updates it to the given state" — implying full overwrite is valid. Any new store implementation following the interface docs will **permanently destroy wallet credentials**.

**Fix:** Either pass the full state, add a separate `UpdateIndex(ctx, uint32)` method, or make the merge contract explicit and mandatory in the interface docs.

---

### 🟠 HIGH

**H1. `discover.go:25-83` — Discovery only checks offchain VTXOs, misses onchain/boarding/redemption-only funded keys**

`discoverHDWalletKeys` derives offchain scripts and queries the indexer for VTXOs. If a key was used exclusively for boarding (user sent BTC but never completed a round) or for redemption/unilateral exit, **those keys won't be discovered on restore**. Funds at boarding-only or redemption-only addresses beyond the current gap window become invisible after wallet restore.

The PR description mentions "boarding UTXOs via explorer lookups" and "redemption/materialized VTXO outputs via explorer lookups" as part of discovery, but the code only scans offchain VTXOs. Either the description is aspirational or implementation is missing.

**H2. `wallet.go:38,133,220` — Mnemonic stored as plaintext Go string indefinitely**

The `service` struct holds `mnemonic string` for the lifetime of the unlocked wallet. Go strings are immutable and cannot be zeroed. Process memory dumps, core dumps, or swap-to-disk expose the master secret.

**Fix:** Store as `[]byte` and zero in `Lock()`. Ideally, don't store the mnemonic at all — re-decrypt from the store on demand when `Dump()` is called.

**H3. `key_service.go:24,109` — Private keys cached by value in map, copies cannot be zeroed**

`derivedKeyCache map[uint32]btcec.PrivateKey` stores keys by value. When `Lock()` nils `keyProvider`, the GC eventually collects memory, but key bytes persist until overwritten.

**Fix:** Store `*btcec.PrivateKey` pointers and explicitly `key.Zero()` each entry on lock. Or avoid caching entirely — HD derivation is fast.

**H4. `wallet.go:84,90,93` — `masterSeed` (64-byte BIP39 seed) not zeroed after use**

The seed can derive the entire key tree. It goes out of scope without zeroing after `Create`.

**Fix:** `defer zeroBytes(masterSeed)` immediately after creation. Same for `entropy` at line 76.

**H5. `init.go:86-91` — `Unlock` modifies `syncDone`, `syncErr`, `syncCh` without holding `syncMu`**

```go
a.syncDone = false
a.syncErr = nil
a.syncCh = make(chan error)
```

These fields are protected by `syncMu` everywhere else (`IsSynced`, `setRestored`, `Lock`, `Reset`, `Stop`, `safeCheck`), but `Unlock` writes them unprotected. Data race if `IsSynced` or `safeCheck` is called concurrently.

**H6. `wallet.go:564-565` — `signTaprootKeySpend` tweaks with `nil`, latent risk for tap-tree outputs**

```go
sig, err := schnorr.Sign(txscript.TweakTaprootPrivKey(*prvkey, nil), preimage)
```

Correct for pure key-path spends, but if any future code path sends a tap-tree-bearing input through this function, the signature will be invalid. Worth adding a guard/assertion that the input has no tap tree commitment.

---

### 🟡 MEDIUM

**M1. `utils.go:32-53` — `parseDerivationIndex` silently discards intermediate path components**

`"m/1/5"` parses to `{0, 5}`, silently ignoring the account `1`. Could lead to wrong-key derivation if different accounts are ever used.

**Fix:** Validate the full path matches `m/{account}/{index}` format exactly.

**M2. `init.go:60-66` — `Init` hardcodes `WalletType: client.SingleKeyWallet`**

Despite the SDK now defaulting to HD wallets, `Init` still passes `SingleKeyWallet` to client-lib. Mismatch could cause subtle bugs if client-lib uses this field for any decision-making.

**M3. `wallet.go:447-478` — `signTapscriptSpend` returns error after first non-matching leaf**

If the first tapscript leaf doesn't contain the owner's pubkey, the function errors immediately instead of trying remaining leaves. Multi-leaf inputs where the owner's key is in a non-first leaf will fail to sign.

**M4. Cross-repo: Dependency on unmerged `arkd#1008`**

`client-lib` is pinned to commit `7ff1fffcb409` which appears to be from an unmerged PR. If that PR is rebased or abandoned, this commit becomes unreachable.

**M5. Cross-repo: `bancod` mock clients lack `Wallet()` method**

The new `Wallet()` on `ArkClient` interface means `bancod`'s `mockArkClient` structs (embedding the interface) will panic at runtime if `Wallet()` is called in any code path.

**M6. `discover.go:78-83` — Discovery allocates keys via N sequential `NewKey` calls, each doing a store write**

For a restored wallet with keys up to index 1000, this means 1000 sequential disk writes. Functional but slow. A batch `SetNextIndex` method would be safer and faster. Also: if the process crashes mid-loop, partial state persists (though re-discovery on next unlock handles this).

---

### 🔵 LOW

- **L1. `wallet.go:261`** — Typo: `keiID` → `keyID`
- **L2. `funding.go:352-356`** — `getIndex` silently returns 0 on parse failure (error from `ParseUint` ignored)
- **L3. `client.go:131-137`** — `syncListeners` allocated twice in constructor, first immediately discarded
- **L4. `key_service.go:128-130`** — Display path `m/0/{index}` omits full BIP86 prefix; could confuse debugging
- **L5. `wallet.go:416-422`** — `NewVtxoTreeSigner` uses ephemeral random key, not HD-derived. Confirm intentional and add comment explaining why.

---

### 🧪 Test Coverage Gaps

1. **No unit test for `signingKeysByScript`** — the critical bridge between HD key IDs and transaction scripts. The positional onchain address lookup (C1) is completely untested.
2. **No e2e test for unilateral exit with HD keys** — `Unroll`, `CompleteUnroll`, `WithdrawFromAllExpiredBoardings` are protocol-critical forfeit/exit paths, untested with HD wallet.
3. **No e2e test for collaborative exit with HD keys** — another protocol-critical path.
4. **No test for wallet locked during send** — TOCTOU window between `getSpendableVtxos` and `signingKeysByScript`.
5. **No concurrent wallet-level test** — `keyService` has concurrency tests, but the `service` wrapper (with its own mutex) does not.
6. **No test for restore with boarding-only funded keys** — would expose H1 (discovery misses non-offchain keys).

---

### Verdict

**Request changes.** Three critical findings (C1-C3) must be addressed before merge. The password hash issue (C2) is the most urgent — it's a standard cryptographic mistake that materially weakens wallet security. The positional KeyID inference (C1) is a latent funds-loss vector. The `persistState` contract (C3) is a correctness timebomb.

The high-severity memory hygiene issues (H2-H4) are standard expectations for a Bitcoin wallet and should be addressed in this PR or a fast follow-up.

⚠️ **Protocol-critical: This PR touches transaction signing and key derivation. Human review and sign-off required before merge.**

ghost (changes requested):
## 🔍 Arkana Code Review — HD Wallet (#143)

This is protocol-critical code (key derivation, transaction signing, fund custody). Thorough review follows. **Human sign-off required before merge per protocol-critical policy.**

---

### 🔴 CRITICAL

**C1. `signing_keys.go:60-75` — Positional index correlation for onchain address KeyIDs is a funds-loss vector**

Onchain addresses have no `KeyID`. The code infers them by matching array indices with offchain/boarding/redemption address lists:

```go
for i, addr := range onchainAddrs {
    keyID := ""
    if i < len(offchainAddrs) {
        keyID = offchainAddrs[i].KeyID
    }
```

This assumes `getAllocatedAddresses` returns all four address types in lock-step order and equal length. If any address type is ever filtered, skipped, or reordered (e.g., by an upstream `GetAddresses` change), **the wrong signing key binds to an onchain UTXO**. A transaction signed with the wrong key is either invalid (funds stuck) or, in collaborative exit paths, could produce unexpected behavior.

**Fix:** Onchain addresses should carry their own `KeyID`, or maintain an explicit `outputScript → keyID` mapping derived from the key itself.

**C2. `wallet.go:108` — Unsalted single-SHA-256 password hash negates PBKDF2 protection**

```go
passwordHash := sha256.Sum256(pwd)
```

The encryption key uses PBKDF2 with 600k iterations (good), but the password *verification* hash is a single unsalted SHA-256. An attacker with the wallet file brute-forces the SHA-256 hash at billions/sec on commodity GPUs to recover the password, then decrypts the master key. The verification hash is the weakest link.

**Fix:** Remove `PasswordHash` entirely. Verify passwords by attempting AES-GCM decryption — if authentication fails, the password is wrong. This is already the behavior in `Unlock` at line 200. The `PasswordHash` field is redundant and harmful.

**C3. `wallet.go:573-586` — `persistState` sends zero-value credentials to store, relying on implicit merge contract**

```go
return w.store.Save(ctx, walletstore.State{
    NextIndex: w.keyProvider.GetNextKeyIndex(),
})
```

`EncryptedMasterKey`, `EncryptedMnemonic`, `PasswordHash` are all zero-valued. Current store implementations (file and in-memory) happen to merge correctly, but the `Store.Save` interface contract says "stores the given state if not yet persisted otherwise updates it to the given state" — implying full overwrite is valid. Any new store implementation following the interface docs will **permanently destroy wallet credentials**.

**Fix:** Either pass the full state, add a separate `UpdateIndex(ctx, uint32)` method, or make the merge contract explicit and mandatory in the interface docs.

---

### 🟠 HIGH

**H1. `discover.go:25-83` — Discovery only checks offchain VTXOs, misses onchain/boarding/redemption-only funded keys**

`discoverHDWalletKeys` derives offchain scripts and queries the indexer for VTXOs. If a key was used exclusively for boarding (user sent BTC but never completed a round) or for redemption/unilateral exit, **those keys won't be discovered on restore**. Funds at boarding-only or redemption-only addresses beyond the current gap window become invisible after wallet restore.

The PR description mentions "boarding UTXOs via explorer lookups" and "redemption/materialized VTXO outputs via explorer lookups" as part of discovery, but the code only scans offchain VTXOs. Either the description is aspirational or implementation is missing.

**H2. `wallet.go:38,133,220` — Mnemonic stored as plaintext Go string indefinitely**

The `service` struct holds `mnemonic string` for the lifetime of the unlocked wallet. Go strings are immutable and cannot be zeroed. Process memory dumps, core dumps, or swap-to-disk expose the master secret.

**Fix:** Store as `[]byte` and zero in `Lock()`. Ideally, don't store the mnemonic at all — re-decrypt from the store on demand when `Dump()` is called.

**H3. `key_service.go:24,109` — Private keys cached by value in map, copies cannot be zeroed**

`derivedKeyCache map[uint32]btcec.PrivateKey` stores keys by value. When `Lock()` nils `keyProvider`, the GC eventually collects memory, but key bytes persist until overwritten.

**Fix:** Store `*btcec.PrivateKey` pointers and explicitly `key.Zero()` each entry on lock. Or avoid caching entirely — HD derivation is fast.

**H4. `wallet.go:84,90,93` — `masterSeed` (64-byte BIP39 seed) not zeroed after use**

The seed can derive the entire key tree. It goes out of scope without zeroing after `Create`.

**Fix:** `defer zeroBytes(masterSeed)` immediately after creation. Same for `entropy` at line 76.

**H5. `init.go:86-91` — `Unlock` modifies `syncDone`, `syncErr`, `syncCh` without holding `syncMu`**

```go
a.syncDone = false
a.syncErr = nil
a.syncCh = make(chan error)
```

These fields are protected by `syncMu` everywhere else (`IsSynced`, `setRestored`, `Lock`, `Reset`, `Stop`, `safeCheck`), but `Unlock` writes them unprotected. Data race if `IsSynced` or `safeCheck` is called concurrently.

**H6. `wallet.go:564-565` — `signTaprootKeySpend` tweaks with `nil`, latent risk for tap-tree outputs**

```go
sig, err := schnorr.Sign(txscript.TweakTaprootPrivKey(*prvkey, nil), preimage)
```

Correct for pure key-path spends, but if any future code path sends a tap-tree-bearing input through this function, the signature will be invalid. Worth adding a guard/assertion that the input has no tap tree commitment.

---

### 🟡 MEDIUM

**M1. `utils.go:32-53` — `parseDerivationIndex` silently discards intermediate path components**

`"m/1/5"` parses to `{0, 5}`, silently ignoring the account `1`. Could lead to wrong-key derivation if different accounts are ever used.

**Fix:** Validate the full path matches `m/{account}/{index}` format exactly.

**M2. `init.go:60-66` — `Init` hardcodes `WalletType: client.SingleKeyWallet`**

Despite the SDK now defaulting to HD wallets, `Init` still passes `SingleKeyWallet` to client-lib. Mismatch could cause subtle bugs if client-lib uses this field for any decision-making.

**M3. `wallet.go:447-478` — `signTapscriptSpend` returns error after first non-matching leaf**

If the first tapscript leaf doesn't contain the owner's pubkey, the function errors immediately instead of trying remaining leaves. Multi-leaf inputs where the owner's key is in a non-first leaf will fail to sign.

**M4. Cross-repo: Dependency on unmerged `arkd#1008`**

`client-lib` is pinned to commit `7ff1fffcb409` which appears to be from an unmerged PR. If that PR is rebased or abandoned, this commit becomes unreachable.

**M5. Cross-repo: `bancod` mock clients lack `Wallet()` method**

The new `Wallet()` on `ArkClient` interface means `bancod`'s `mockArkClient` structs (embedding the interface) will panic at runtime if `Wallet()` is called in any code path.

**M6. `discover.go:78-83` — Discovery allocates keys via N sequential `NewKey` calls, each doing a store write**

For a restored wallet with keys up to index 1000, this means 1000 sequential disk writes. Functional but slow. A batch `SetNextIndex` method would be safer and faster. Also: if the process crashes mid-loop, partial state persists (though re-discovery on next unlock handles this).

---

### 🔵 LOW

- **L1. `wallet.go:261`** — Typo: `keiID` → `keyID`
- **L2. `funding.go:352-356`** — `getIndex` silently returns 0 on parse failure (error from `ParseUint` ignored)
- **L3. `client.go:131-137`** — `syncListeners` allocated twice in constructor, first immediately discarded
- **L4. `key_service.go:128-130`** — Display path `m/0/{index}` omits full BIP86 prefix; could confuse debugging
- **L5. `wallet.go:416-422`** — `NewVtxoTreeSigner` uses ephemeral random key, not HD-derived. Confirm intentional and add comment explaining why.

---

### 🧪 Test Coverage Gaps

1. **No unit test for `signingKeysByScript`** — the critical bridge between HD key IDs and transaction scripts. The positional onchain address lookup (C1) is completely untested.
2. **No e2e test for unilateral exit with HD keys** — `Unroll`, `CompleteUnroll`, `WithdrawFromAllExpiredBoardings` are protocol-critical forfeit/exit paths, untested with HD wallet.
3. **No e2e test for collaborative exit with HD keys** — another protocol-critical path.
4. **No test for wallet locked during send** — TOCTOU window between `getSpendableVtxos` and `signingKeysByScript`.
5. **No concurrent wallet-level test** — `keyService` has concurrency tests, but the `service` wrapper (with its own mutex) does not.
6. **No test for restore with boarding-only funded keys** — would expose H1 (discovery misses non-offchain keys).

---

### Verdict

**Request changes.** Three critical findings (C1-C3) must be addressed before merge. The password hash issue (C2) is the most urgent — it's a standard cryptographic mistake that materially weakens wallet security. The positional KeyID inference (C1) is a latent funds-loss vector. The `persistState` contract (C3) is a correctness timebomb.

The high-severity memory hygiene issues (H2-H4) are standard expectations for a Bitcoin wallet and should be addressed in this PR or a fast follow-up.

⚠️ **Protocol-critical: This PR touches transaction signing and key derivation. Human review and sign-off required before merge.**

wallet/hdwallet/key_service.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+const (
+	defaultAccount = uint32(0)
+)
+
+// keyService handles BIP32 key derivation and caching for HD wallets.
+type keyService struct {
+	// masterKey is the BIP32 root key restored from the user's mnemonic or xpriv.
+	masterKey *hdkeychain.ExtendedKey
```
sekulicd, line 20: I think the BIP86 account-root model is OK, but the current implementation hides too much context and uses misleading names.

Right now `Create` derives:

```text
m/86'/coin_type'/0'
```

and stores that derived account xpriv as `EncryptedMasterKey`.

Then `key_service.go` derives child keys as:

```text
m/0/{index}
```

So the actual absolute path for key `m/0/5` is:

```text
m/86'/coin_type'/0'/0/5
```

The issue is not the derivation itself. The issue is that `EncryptedMasterKey` and `masterKey` imply we are storing the BIP32 root from the seed, but we are actually storing the BIP86 account key. Also, returning `m/0/{index}` as the key ID looks like an absolute derivation path, while it is really account-relative.

Suggested changes:

- Rename `EncryptedMasterKey` to `EncryptedAccountKey` or `EncryptedAccountXPriv`.
- Rename `keyService.masterKey` to `accountKey` if it stores `m/86'/coin_type'/0'`.
- Add comments documenting:
  ```text
  BIP39 seed -> BIP32 root -> account key m/86'/coin_type'/0'
  stored xpriv = account key
  child key IDs are currently account-relative m/0/{index}
  absolute child path = m/86'/coin_type'/0'/0/{index}
  ```
- Prefer returning full absolute key IDs:
  ```text
  m/86'/coin_type'/0'/0/{index}
  ```
  even if internally we store the account xpriv and only derive the relative suffix.
- If we keep relative key IDs, make that explicit in naming/comments and avoid calling them generic derivation paths.
- Tighten `parseDerivationIndex`: paths like `m/1/5` should not silently resolve to the same key as `m/0/5`. If only account-relative `m/0/{index}` is supported, reject other branches.

This keeps the BIP86 account-root approach but removes ambiguity for debugging, restore behavior, cross-SDK checks, and future migrations.
```

altafan:
@arkanaai review again

Issue #150, opened 2026-04-22
VHTLC contract


altafan:
Let's add support for VHTLC contract.

Ref: https://github.com/arkade-os/ts-sdk/blob/master/src/contracts/handlers/vhtlc.ts

tiero:
Do we really need a bespoke contract type for it?

I guess we could just treat it as we would treat the other custom script contracts, and ideally we expose this as template, but since go doesnt really have a concept of class/inheritance I would just have a file with the opcodes to encode [to a script, like we do in this "example"](https://docs.arkadeos.com/contracts/hashlock)

for spending we can always have a dedicated pkg that knows how to craft the custom transactions (claim, refunds), in meantime we progress with the compiler work so we could statically generate these custom tx building, starting from a contract template

wdyt?

bitcoin-coder-bob:
working off PR: https://github.com/arkade-os/go-sdk/pull/157

PR #157, opened 2026-04-28
WIP: Feat/vhtlc contract

bitcoin-coder-bob:
closes https://github.com/arkade-os/go-sdk/issues/150

## VHTLC contract

Adds Virtual HTLC support on top of the delegate PR (#156).

- `vhtlc/`: a small package (`vhtlc.go`, `opts.go`, `utils.go`) wrapping the VHTLC tapscript construction (preimage hash, sender/receiver/server keys, refund and claim leaves). Lifted from the existing inline construction so the contract handler can reuse it.
- `contract/vhtlc_handler.go`, `contract/vhtlc_handler_test.go`: new `VHTLCHandler` deriving an offchain VHTLC contract; `SelectPath`/`GetSpendablePaths` expose claim, refund (with delay), and cooperative leaves to callers. Tests cover deterministic derivation, param round-trip, missing-param errors, path selection, and spendable-path enumeration.
- `contract/manager.go`, `contract/types.go`: `VHTLCCreator` interface with `NewVHTLC(ctx, rawParams)`; `contractManager.NewVHTLC` derives via `VHTLCHandler`, persists, and emits a `contract_created` event under the same closure+defer mutex pattern used by `NewDelegate`.
- `types/types.go`: added `ContractTypeVHTLC` constant and `Contract.GetTapscripts()` helper that decodes the JSON-encoded `Params["tapscripts"]` list (used by the watcher and signing paths).

## Stacked on #156

Base is `feat/delegate-contract`. This PR's diff against `master` therefore includes the watcher (#153) and delegate (#156) work. Land those first.

## Status

WIP, still iterating. Currently green on `go build ./...`, `make lint` (0 issues), and the full unit test suite under `-race`.


ghost (changes requested):
## 🔴 Aggressive Code Review — arkade-os/go-sdk#157 "Feat/vhtlc contract"

Thorough review of the full diff (+5954/−273 across 36 files). This PR introduces a contract manager abstraction, a watcher, a delegate handler, and a VHTLC handler. It's a major architectural change that touches protocol-critical paths. Findings below, ordered by severity.

---

### ⚠️ PROTOCOL-CRITICAL — Requires human sign-off regardless of fixes

This PR modifies VTXO handling, spend path selection, forfeit paths, and transaction signing flows. Even if all findings below are resolved, a human must review and approve before merge per the protocol-critical rules.

---

### 🔴 P0 — Data races on `a.contractManager` (funds at risk)

**`client.go` ~line 1485 (`safeCheck`), ~line 1501 (`getAllBoardingUtxos`), ~line 1741 (`saveSendTransaction`), `send.go` (`getSpendableVtxos`), `funding.go` (`NewOffchainAddress`, `NewBoardingAddress`, `NewOnchainAddress`):**

All these methods read `a.contractManager` directly **without holding `cmMu`**. Meanwhile, `Lock()`, `Reset()`, and `Stop()` set `a.contractManager = nil` under `cmMu.Lock()`.

The `contractMgr()` accessor exists and correctly uses `cmMu.RLock()`, but it's only used in `ContractManager()` and `listenForArkTxs`. Every other call site bypasses it.

**TOCTOU scenario:** `safeCheck()` observes `a.contractManager != nil` → concurrent `Lock()` sets it to `nil` → caller dereferences a nil pointer → panic or silent data corruption.

```go
// safeCheck reads without lock:
if a.contractManager == nil {  // ← DATA RACE
    return ErrNotInitialized
}
// ... caller then uses a.contractManager without lock
```

**Fix:** All reads of `a.contractManager` must go through `contractMgr()` or hold `cmMu.RLock()`. Alternatively, `safeCheck()` should return the manager under lock so the caller can use the returned value safely.

---

### 🔴 P0 — `serializeVHTLCOpts` silently swallows BIP68 errors (timelock bypass)

**`contract/vhtlc_handler.go` ~lines 151-154 (`serializeVHTLCOpts`):**

```go
claimSeq, _ := arklib.BIP68Sequence(opts.UnilateralClaimDelay)
refundSeq, _ := arklib.BIP68Sequence(opts.UnilateralRefundDelay)
noRcvrSeq, _ := arklib.BIP68Sequence(opts.UnilateralRefundWithoutReceiverDelay)
```

If `BIP68Sequence` fails (e.g. invalid locktime type), the sequence is silently set to `0`. A contract serialized with sequence `0` has **no CSV timelock**, meaning the unilateral exit path could be spent immediately. This defeats the entire security model of the VHTLC.

**Fix:** Propagate the error. `serializeVHTLCOpts` should return `(map[string]string, error)`.

---

### 🔴 P1 — Public API break: `GetAddresses` return type change

**`sdk.go` ~line 34:**

```go
// Before:
GetAddresses(ctx context.Context) (
    onchainAddresses, offchainAddresses, boardingAddresses, redemptionAddresses []string, err error)

// After:
GetAddresses(ctx context.Context) (
    onchainAddresses []string,
    offchainAddresses, boardingAddresses, redemptionAddresses []clientTypes.Address, err error)
```

This breaks all downstream consumers of the `ArkClient` interface. Confirmed consumers:
- `bancod/pkg/contract/taker.go:103` — `_, offchainAddrs, _, _, err := wp.Wallet().GetAddresses(ctx)`
- `demos/asset/golang/main.go:100`
- `asset-demos/asset/golang/main.go:100`

These repos compile against go-sdk and will fail. Need coordinated PRs on downstream repos or a migration path.

---

### 🟡 P2 — `NewVHTLC` TOCTOU: duplicate event emission

**`contract/manager.go` ~lines 1976-1998 (`NewVHTLC`):**

```go
m.mu.RLock()
existing, found := m.contracts[c.Script]
m.mu.RUnlock()
if found { return &existing, nil }
// ← Another goroutine can reach here with same script
c.CreatedAt = time.Now()
if err := m.persistAndCache(ctx, *c); err != nil { ... }
m.emit(Event{Type: "contract_created", Contract: *c})
```

Unlike `NewDefault` and `NewDelegate` which hold dedicated create mutexes, `NewVHTLC` has no serialization. Two concurrent calls with identical params will both pass the existence check, both persist (benign due to upsert), and both emit `contract_created` events. The watcher will attempt to subscribe the same address twice.

**Fix:** Add a `vhtlcCreateMu` or use the script as the lock key.

---

### 🟡 P2 — Deprecated `ripemd160` import

**`vhtlc/vhtlc.go` line 11:**

```go
"golang.org/x/crypto/ripemd160" //nolint:staticcheck
```

`btcutil.Hash160` (used in the tests) wraps the same logic. Using the deprecated package directly is unnecessary. Use `btcutil.Hash160` in production code for consistency.

---

### 🟡 P2 — Inconsistent pubkey serialization: delegate key is 33 bytes, all others are 32

**`contract/delegate_handler.go` ~line 149:**

```go
ParamDelegateKey: hex.EncodeToString(delegateKey.SerializeCompressed()),
```

Every other pubkey param uses `schnorr.SerializePubKey` (32-byte x-only). The delegate key uses `SerializeCompressed()` (33-byte SEC). This inconsistency will cause confusion if any downstream code tries to deserialize it as a schnorr key or compare it against other params.

If the 33-byte form is intentional (e.g. to preserve the y-coordinate parity), add a comment explaining why. Otherwise, standardize on 32-byte x-only.

---

### 🟡 P2 — `ListContracts` SQL filter ignores `IsOnchain` and `KeyID`

**`store/sql/contract_store.go` `ListContracts`:**

The `Filter` struct has `IsOnchain` and `KeyID` fields, but `ListContracts` only applies `Type`, `State`, and `Script` filters. The `IsOnchain` and `KeyID` filters are silently ignored at the SQL layer (they only work in the in-memory `GetContracts` path in `manager.go`).

This means querying the persistent store directly (e.g. during `Load()`) won't honor these filters. Currently `Load()` always passes `Filter{}` so it's not a bug today, but it's a trap for future callers.

**Fix:** Add `IsOnchain` and `KeyID` (via JSON extract) conditions to the SQL query.

---

### 🟢 P3 — Nit: `toBitcoinNetwork` duplicated → now returns error instead of defaulting to mainnet

**`contract/default_handler.go` vs old `utils.go`:**

The old `toBitcoinNetwork` silently defaulted to mainnet for unknown networks. The new version returns an error. This is **good** — it's safer. Just noting the behavior change for awareness.

---

### 🟢 P3 — `Opts()` extracts preimage hash via slice offset `[2:2+hash160Len]`

**`vhtlc/vhtlc.go` `Opts()` method:**

```go
PreimageHash: v.preimageConditionScript[2 : 2+hash160Len],
```

This assumes the condition script layout is always `OP_HASH160 <20-byte push> OP_EQUAL` with the hash at offset 2. If the script builder ever changes the push encoding (e.g. for longer hashes), this will silently return wrong bytes. Consider parsing the script properly or at minimum adding a length assertion.

---

### 🟢 P3 — Good: VHTLC tapscript leaf ordering matches ts-sdk

Verified against `/root/arkana/repos/ts-sdk/src/script/vhtlc.ts` lines 110-117. The 6-leaf ordering is an exact match. Cross-SDK VHTLC interoperability is preserved. ✅

---

### 🟢 P3 — Good: Delegate closure ordering test

`TestDelegateHandler_ClosureOrdering` correctly verifies that `forfeitClosures[0]` is the 2-of-2 (owner+server) and not the 3-of-3 (owner+delegate+server). This is important because arkd uses `forfeitClosures[0]` for forfeit tx construction. ✅

---

### 🟢 P3 — Good: Test coverage

2,800+ lines of new tests covering:
- Handler derivation determinism
- Path selection for all contract types
- Manager concurrency (20-goroutine test for NewDelegate)
- Watcher lifecycle (subscribe, backoff, reconnect, dedup, stop)
- SQL store CRUD and filter round-trips
- VHTLC param round-trip serialization

---

### Summary

| Severity | Count | Key Issues |
|----------|-------|------------|
| 🔴 P0 | 2 | Data races on contractManager; silent BIP68 error swallowing |
| 🔴 P1 | 1 | Public API break (GetAddresses) with known downstream consumers |
| 🟡 P2 | 3 | VHTLC TOCTOU, deprecated ripemd160, SQL filter gap |
| 🟢 P3 | 4 | Nits, observations, positive callouts |

**Requesting changes** for the P0 data race and the P0 BIP68 silent error. The P1 API break needs a migration plan with downstream repos (bancod, demos).

The VHTLC tapscript layout, closure ordering, and cross-SDK compatibility all look correct. Architecture is solid. Fix the safety issues and this is ready for human sign-off.

ghost (approved):
## ✅ Follow-up Review #3 — arkade-os/go-sdk#157 (commit `73cfa24d`)

Reviewed diff `a1c38fd8..73cfa24d` ("fix vhtlc security review findings").

### All previous findings resolved

| Previous Finding | Status |
|---|---|
| 🔴 P0 `serializeVHTLCOpts` BIP68 error swallowing | ✅ Fixed — returns `(map[string]string, error)`, all 3 BIP68 errors propagated (`vhtlc_handler.go:337-349`) |
| 🟡 P2 `GetSpendablePaths` missing CLTV check | ✅ Fixed — now gates on `isVHTLCCltvSatisfied` (`vhtlc_handler.go:216`), matching `SelectPath` behavior |
| 🟡 P2 `NewVHTLC` TOCTOU race | ✅ Fixed — `m.mu.Lock()` now wraps check-then-insert (`manager.go:334-353`), unlock before `emit` is intentional and correct |
| 🟡 P2 `SelectPath` swallowed `vhtlcRefundLocktime` error | ✅ Fixed — error propagated (`vhtlc_handler.go:154-156`) |
| 🟡 P2 `GetSpendablePaths` swallowed `vhtlcRefundLocktime` error | ✅ Fixed — error propagated (`vhtlc_handler.go:217-219`) |
| 🟢 P3 `GetRevealedTapscripts` silent error drop | ✅ Fixed — now returns `([]string, error)` (`vhtlc.go:172`), caller checks error (`vhtlc_handler.go:88-90`) |

### Tests updated
- `GetSpendablePaths` sender collaborative test split into CLTV-satisfied and CLTV-not-satisfied cases (`vhtlc_handler_test.go:413-441`)

### No new issues found

Code is clean. Approving.

⚠️ **PROTOCOL-CRITICAL**: This PR adds VHTLC contract construction and spend path selection (hash-locked claims, timelocked refunds, unilateral exits). **Requires human sign-off before merge.**

---
*Reviewed by Arkana 🔮 — 3rd pass, focused on commit `73cfa24d`*

ghost (commented):
## 🔮 Arkana Review — 4th pass (merge commit `d77095a6`)

Focused on delta since last review (`73cfa24d` → `d77095a6`). This merge brings in `feat/delegate-contract` changes.

### Changes reviewed

1. **`contract/types.go` — `DelegateCreator` interface extracted from `Manager`**
2. **`contract/manager.go` — Dedup logic in `NewDelegate` + `findDelegateContractByKey` helper**
3. **`contract/delegate_handler.go` — Nil checks for owner/signer keys**

---

### ✅ Cross-repo impact: CLEAR

Searched all repos under `/root/arkana/repos/` — no external consumer imports `go-sdk/contract` or references `contract.Manager` directly. The `NewDelegate` → `DelegateCreator` extraction is safe. `bancod`, `introspector-review`, `demos` all consume `go-sdk` at the top-level `ArkClient` level only.

### ✅ Interface compliance: CLEAR

Both test mocks (`watcherMockManager`, `fixContractManager`) implement `NewVHTLC` to satisfy the updated `Manager` interface. Extra `NewDelegate` methods on mocks are harmless in Go.

### ✅ Nil checks in `delegate_handler.go:84-89`: GOOD

Defensive checks for `key.PubKey` and `cfg.SignerKey` before use. Prevents nil-pointer panics.

### ⚠️ Minor: Manual mutex in `NewDelegate` (`manager.go:250-312`)

`NewDelegate` has **9 separate `m.mu.Unlock()` calls** across error paths. All paths are correctly covered today, but this is fragile — future edits could easily miss an unlock. Consider refactoring to use a closure pattern:

```go
result, err := func() (*types.Contract, error) {
    m.mu.Lock()
    defer m.mu.Unlock()
    // ... critical section
}()
```

Not blocking, but worth hardening.

### ✅ Dedup logic in `findDelegateContractByKey` (`manager.go:565-577`): CORRECT

Linear scan over delegate contracts by type. O(n) per call, called under mutex. Acceptable for current scale. If delegate contract count grows significantly, consider indexing by key in the store layer.

### ✅ CI: Passing

---

**Verdict:** Merge commit is clean. No protocol-critical changes, no security issues, no breakage. Previous review findings on the VHTLC code still apply.

*Reviewed by Arkana 🔮 — 4th pass, focused on merge commit `d77095a6`*

bitcoin-coder-bob:
> ## 🔮 Arkana Review — 4th pass (merge commit `d77095a6`)
> Focused on delta since last review (`73cfa24d` → `d77095a6`). This merge brings in `feat/delegate-contract` changes.
> 
> ### Changes reviewed
> 1. **`contract/types.go` — `DelegateCreator` interface extracted from `Manager`**
> 2. **`contract/manager.go` — Dedup logic in `NewDelegate` + `findDelegateContractByKey` helper**
> 3. **`contract/delegate_handler.go` — Nil checks for owner/signer keys**
> 
> ### ✅ Cross-repo impact: CLEAR
> Searched all repos under `/root/arkana/repos/` — no external consumer imports `go-sdk/contract` or references `contract.Manager` directly. The `NewDelegate` → `DelegateCreator` extraction is safe. `bancod`, `introspector-review`, `demos` all consume `go-sdk` at the top-level `ArkClient` level only.
> 
> ### ✅ Interface compliance: CLEAR
> Both test mocks (`watcherMockManager`, `fixContractManager`) implement `NewVHTLC` to satisfy the updated `Manager` interface. Extra `NewDelegate` methods on mocks are harmless in Go.
> 
> ### ✅ Nil checks in `delegate_handler.go:84-89`: GOOD
> Defensive checks for `key.PubKey` and `cfg.SignerKey` before use. Prevents nil-pointer panics.
> 
> ### ⚠️ Minor: Manual mutex in `NewDelegate` (`manager.go:250-312`)
> `NewDelegate` has **9 separate `m.mu.Unlock()` calls** across error paths. All paths are correctly covered today, but this is fragile — future edits could easily miss an unlock. Consider refactoring to use a closure pattern:
> 
> ```go
> result, err := func() (*types.Contract, error) {
>     m.mu.Lock()
>     defer m.mu.Unlock()
>     // ... critical section
> }()
> ```
> 
> Not blocking, but worth hardening.
> 
> ### ✅ Dedup logic in `findDelegateContractByKey` (`manager.go:565-577`): CORRECT
> Linear scan over delegate contracts by type. O(n) per call, called under mutex. Acceptable for current scale. If delegate contract count grows significantly, consider indexing by key in the store layer.
> 
> ### ✅ CI: Passing
> **Verdict:** Merge commit is clean. No protocol-critical changes, no security issues, no breakage. Previous review findings on the VHTLC code still apply.
> 
> _Reviewed by Arkana 🔮 — 4th pass, focused on merge commit `d77095a6`_

"Manual mutex in `NewDelegate`" addressed in [b837270](https://github.com/arkade-os/go-sdk/pull/157/commits/b837270863e6824a4c912de3af1c9cb04fbff07f)

ghost (commented):
## 🔮 Arkana Review — 5th pass (commit `b8372708`)

Focused on delta since last review (`d77095a6` → `b8372708`).

### ✅ NewDelegate mutex refactor — looks good

The closure+`defer m.mu.Unlock()` pattern in `NewDelegate` is correct and addresses the manual-unlock feedback from pass 4. The `isNew` guard on `m.emit()` is a nice touch — avoids emitting for already-existing contracts returned from the cache path.

### ⚠️ One remaining nit: NewVHTLC still uses manual unlocks

`contract/manager.go:NewVHTLC` (~line 370–400) still does manual `m.mu.Unlock()` on each error path — the same pattern that was just cleaned up in `NewDelegate`. For consistency and safety, consider applying the same closure+defer treatment. Not a blocker, but it'd be inconsistent to fix one and leave the other.

No new protocol-correctness or security issues found in this delta.

*Reviewed by Arkana 🔮 — 5th pass, focused on commit `b8372708`*

ghost (commented):
## 🔮 Arkana Review — 5th pass (commit `b8372708`)

Focused on delta since last review (`d77095a6` → `b8372708`).

### ✅ NewDelegate mutex refactor — looks good

The closure+`defer m.mu.Unlock()` pattern in `NewDelegate` is correct and addresses the manual-unlock feedback from pass 4. The `isNew` guard on `m.emit()` is a nice touch — avoids emitting for already-existing contracts returned from the cache path.

### ⚠️ One remaining nit: NewVHTLC still uses manual unlocks

`contract/manager.go:NewVHTLC` (~line 370–400) still does manual `m.mu.Unlock()` on each error path — the same pattern that was just cleaned up in `NewDelegate`. For consistency and safety, consider applying the same closure+defer treatment. Not a blocker, but it'd be inconsistent to fix one and leave the other.

No new protocol-correctness or security issues found in this delta.

*Reviewed by Arkana 🔮 — 5th pass, focused on commit `b8372708`*

ghost (commented):
## 🔮 Arkana Review — 6th pass (commit `d0bd113`)

Focused on delta since last review (`b8372708` → `d0bd113`).

### Single commit: `d0bd113` — closure+defer mutex pattern for NewVHTLC

Mechanical refactor only. Changes:
- `NewVHTLC` mutex handling moved into an anonymous closure with `defer m.mu.Unlock()`, matching the pattern already used in `NewDelegate` since `b8372708`.
- No logic changes. The lock/unlock boundaries, error paths, return values, and `m.emit()` placement (outside the lock) are all preserved identically.

### Verdict: ✅ Clean

No new findings. The refactor eliminates the risk of forgetting an `m.mu.Unlock()` on a new error path — strictly safer than the previous manual unlock style.

Previous findings from passes 1–5 still apply where not yet addressed.

---
*Reviewed by Arkana 🔮 — 6th pass, focused on commit `d0bd113`*

ghost (commented):
## 🔮 Arkana Review — 6th pass (commit `d0bd113`)

Focused on delta since last review (`b8372708` → `d0bd113`).

### Single commit: `d0bd113` — closure+defer mutex pattern for NewVHTLC

Mechanical refactor only. Changes:
- `NewVHTLC` mutex handling moved into an anonymous closure with `defer m.mu.Unlock()`, matching the pattern already used in `NewDelegate` since `b8372708`.
- No logic changes. The lock/unlock boundaries, error paths, return values, and `m.emit()` placement (outside the lock) are all preserved identically.

### Verdict: ✅ Clean

No new findings. The refactor eliminates the risk of forgetting an `m.mu.Unlock()` on a new error path — strictly safer than the previous manual unlock style.

Previous findings from passes 1–5 still apply where not yet addressed.

---
*Reviewed by Arkana 🔮 — 6th pass, focused on commit `d0bd113`*

PR #158, opened 2026-04-29
fix: graceful shutdown of bg goroutines before DB close

### LINKED ISSUE #154: DB close race: goroutines still running cause sql: database is closed errors
Description
During shutdown, the database connection is closed before all background goroutines have finished their work. As a result, in-flight operations continue to use the DB and fail with errors like:
time="2026-04-23T11:57:47+02:00" level=error msg="failed to process commitment tx" error="sql: database is closed"
This indicates a race condition between DB shutdown and ongoing async processing (e.g. commitment tx handling).
Expected behavior
All goroutines that depend on the DB should gracefully finish (or be canceled) before the database is closed.

Dunsin-cyber:
fixes #154

Added a `goroutineWg sync.WaitGroup` to arkClient to track the 4 background goroutines `(listenForArkTxs, listenForOnchainTxs, listenDbEvents, periodicRefreshDb). Add(4)` is called before the outer goroutine in `Unlock()` so the counter is set before any shutdown can race it. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown synchronization to ensure background processes properly terminate before lock/unlock transitions complete, enhancing reliability and stability.

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,6 +51,7 @@ type arkClient struct {
 	syncErr           error
 	syncListeners     *syncListeners
 	stopFn            context.CancelFunc
+	goroutineWg       sync.WaitGroup
 	stopOnce          sync.Once
 	refreshDbInterval time.Duration
 	dbMu              *sync.Mutex
@@ -267,6 +268,7 @@ func (a *arkClient) Reset(ctx context.Context) {
 
 	if a.stopFn != nil {
 		a.stopFn()
+		a.goroutineWg.Wait()
 	}
 	if a.syncListeners != nil {
 		a.syncListeners.broadcast(fmt.Errorf("wallet reset while restoring"))
@@ -290,6 +292,7 @@ func (a *arkClient) Stop() {
 
 		if a.stopFn != nil {
 			a.stopFn()
+			a.goroutineWg.Wait()
 		}
 		if a.syncListeners != nil {
 			a.syncListeners.broadcast(fmt.Errorf("service stopped while restoring"))
@@ -697,6 +700,7 @@ func groupSpentVtxosByTx(
 }
 
 func (a *arkClient) listenForArkTxs(ctx context.Context) {
+	defer a.goroutineWg.Done()
 	wallet := a.Wallet()
 	if wallet == nil {
 		// Should be unreachable
@@ -778,6 +782,7 @@ func (a *arkClient) listenForArkTxs(ctx context.Context) {
 }
 
 func (a *arkClient) listenForOnchainTxs(ctx context.Context) {
+	defer a.goroutineWg.Done()
 	wallet := a.Wallet()
 	if wallet == nil {
 		// Should be unreachable
@@ -1081,6 +1086,7 @@ func (a *arkClient) listenForOnchainTxs(ctx context.Context) {
 }
 
 func (a *arkClient) listenDbEvents(ctx context.Context) {
+	defer a.goroutineWg.Done()
 	for {
 		select {
 		case <-ctx.Done():
@@ -1139,6 +1145,7 @@ func (a *arkClient) listenDbEvents(ctx context.Context) {
 }
 
 func (a *arkClient) periodicRefreshDb(ctx context.Context) {
+	defer a.goroutineWg.Done()
 	if a.refreshDbInterval == 0 {
 		return
 	}
--- a/init.go
+++ b/init.go
@@ -101,6 +101,7 @@ func (a *arkClient) Unlock(ctx context.Context, password string) error {
 	bgCtx, cancel := context.WithCancel(context.Background())
 	a.stopFn = cancel
 
+	a.goroutineWg.Add(4)
 	go func() {
 		a.Explorer().Start()
 
@@ -136,6 +137,7 @@ func (a *arkClient) Lock(ctx context.Context) error {
 
 	if a.stopFn != nil {
 		a.stopFn()
+		a.goroutineWg.Wait()
 	}
 	if a.syncListeners != nil {
 		a.syncListeners.broadcast(fmt.Errorf("wallet locked while restoring"))
--- a/init.go
+++ b/init.go
@@ -99,10 +99,11 @@ func (a *arkClient) Unlock(ctx context.Context, password string) error {
 	}()
 
 	bgCtx, cancel := context.WithCancel(context.Background())
+	a.goroutineWg.Add(1)
 	a.stopFn = cancel
-
-	a.goroutineWg.Add(4)
 	go func() {
+		defer a.goroutineWg.Done()
+
 		a.Explorer().Start()
 
 		ctx := bgCtx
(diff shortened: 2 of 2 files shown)


ghost (changes requested):
## Arkana Code Review — arkade-os/go-sdk#158

The intent is correct: wait for background goroutines to drain before closing the DB. However, the current implementation has a **critical deadlock bug** and a **silent hang risk**. Requesting changes.

---

### 🔴 CRITICAL: Deadlock when outer goroutine exits early

**`init.go:101`** — `a.goroutineWg.Add(4)` is called unconditionally before the outer goroutine, but the 4 inner goroutines are only spawned at lines 123-128, *after* `discoverHDWalletKeys`, `finalizePendingTxs`, and `refreshDb` all succeed.

If `discoverHDWalletKeys` fails (line 106-110), the outer goroutine returns early. The 4 child goroutines are **never launched**, so `Done()` is never called 4 times. Any subsequent call to `Stop()`, `Lock()`, or `Reset()` will call `goroutineWg.Wait()` and **deadlock forever**.

```go
// init.go — current code in PR
a.goroutineWg.Add(4)       // counter = 4
go func() {
    // ...
    if _, err := a.discoverHDWalletKeys(ctx); err != nil {
        a.syncCh <- err
        close(a.syncCh)
        return              // ← EARLY EXIT: 4 Done()s never happen → DEADLOCK
    }
    // ... only here are the 4 goroutines spawned
}()
```

**Fix:** Move `Add(4)` to just before the 4 `go` calls (line 122), or add compensating `Done()` calls on every early-return path. The cleaner approach:

```go
go func() {
    a.Explorer().Start()
    // ...
    if _, err := a.discoverHDWalletKeys(ctx); err != nil {
        a.syncCh <- err
        close(a.syncCh)
        return
    }
    // ...
    a.goroutineWg.Add(4)  // ← move here, right before spawning
    go a.listenForArkTxs(ctx)
    go a.listenForOnchainTxs(ctx)
    go a.listenDbEvents(ctx)
    go a.periodicRefreshDb(ctx)
}()
```

### 🔴 HIGH: Shutdown during init phase also deadlocks

Even if `discoverHDWalletKeys` succeeds, if `Lock()`/`Stop()`/`Reset()` is called while the outer goroutine is still running `Explorer().Start()`, `discoverHDWalletKeys()`, or `refreshDb()` — i.e. before the 4 goroutines are launched — `goroutineWg.Wait()` will still deadlock because `Add(4)` already ran but no `Done()` has been called yet.

This is the same root cause as above: the `Add` and the `go` launches are not atomic and can be interrupted by shutdown.

Moving `Add(4)` to just before the goroutine launches (as shown above) fixes both issues, because if shutdown happens before `Add(4)`, the WaitGroup is still at 0 and `Wait()` returns immediately.

### 🟡 MEDIUM: No timeout / safety net on `Wait()`

If any of the 4 goroutines hangs (e.g. a blocking gRPC stream that doesn't respect context cancellation), `goroutineWg.Wait()` will block the caller forever. Consider adding a timeout:

```go
done := make(chan struct{})
go func() {
    a.goroutineWg.Wait()
    close(done)
}()
select {
case <-done:
case <-time.After(10 * time.Second):
    log.Warn("timed out waiting for background goroutines to stop")
}
```

### 🟡 MEDIUM: Missing test coverage

There are no tests for:
1. `Stop()` / `Lock()` when `discoverHDWalletKeys` fails (would catch the deadlock)
2. `Stop()` called during the init phase (before goroutines are launched)
3. `Unlock()` → `Lock()` → `Unlock()` cycle (WaitGroup reuse correctness)

Given this is fixing a race condition, at minimum a test that calls `Stop()` shortly after `Unlock()` (to exercise the early-shutdown path) would be valuable.

### ✅ Good

- `defer a.goroutineWg.Done()` at the top of each goroutine is correct — covers early returns (e.g. `wallet == nil` checks, `refreshDbInterval == 0`).
- All three shutdown paths (`Stop`, `Lock`, `Reset`) are covered.
- The `stopOnce` in `Stop()` ensures `Wait()` isn't called concurrently from multiple callers.

---

**Summary:** The approach is sound but the placement of `Add(4)` creates a guaranteed deadlock on any early-exit path in the outer goroutine. Move `Add(4)` to immediately before the 4 `go` calls to fix.

ghost (approved):
## Arkana Code Review — arkade-os/go-sdk#158 (re-review)

Commit `5307875` addresses both critical findings from my previous review:

1. **🔴 Deadlock on early exit** — FIXED. `Add(4)` moved to right before the 4 goroutine launches, after the `refreshDb` error check. Early return no longer leaves the WaitGroup stranded.

2. **🔴 Shutdown during init** — FIXED. Outer goroutine tracked with `Add(1)` + `defer Done()`. `Wait()` now correctly handles shutdown at any phase — during init, during listeners, or after completion.

3. **Bonus fix:** `if err != nil { return }` prevents launching listener goroutines on `refreshDb` failure — good defensive addition.

The WaitGroup lifecycle is now correct:
- `Add(1)` before outer goroutine → `defer Done()` in outer goroutine
- `Add(4)` only reached on success, right before 4 `go` calls → each defers `Done()`
- All 3 shutdown paths (`Lock`, `Stop`, `Reset`) call `stopFn()` then `Wait()`

My two MEDIUM suggestions (timeout on `Wait()`, test coverage) remain unaddressed but are not blocking. They can be follow-ups.

LGTM. ✅

ghost (approved):
## Arkana Code Review — arkade-os/go-sdk#158 (re-review)

Commit `5307875` addresses both critical findings from my previous review:

1. **🔴 Deadlock on early exit** — FIXED. `Add(4)` moved to right before the 4 goroutine launches, after the `refreshDb` error check. Early return no longer leaves the WaitGroup stranded.

2. **🔴 Shutdown during init** — FIXED. Outer goroutine tracked with `Add(1)` + `defer Done()`. `Wait()` now correctly handles shutdown at any phase — during init, during listeners, or after completion.

3. **Bonus fix:** `if err != nil { return }` prevents launching listener goroutines on `refreshDb` failure — good defensive addition.

The WaitGroup lifecycle is now correct:
- `Add(1)` before outer goroutine → `defer Done()` in outer goroutine
- `Add(4)` only reached on success, right before 4 `go` calls → each defers `Done()`
- All 3 shutdown paths (`Lock`, `Stop`, `Reset`) call `stopFn()` then `Wait()`

My two MEDIUM suggestions (timeout on `Wait()`, test coverage) remain unaddressed but are not blocking. They can be follow-ups.

LGTM. ✅

altafan:
@Dunsin-cyber in #153 we're gonna move these listeners into a contract watcher, and therefore this code is going to subject to breaking changes (don't look at it now as it requires indeed changes). If we don't solve directly there, would be better to postpone these fixes after that pr gets merged.

Dunsin-cyber:
> @Dunsin-cyber in #153 we're gonna move these listeners into a contract watcher, and therefore this code is going to subject to breaking changes (don't look at it now as it requires indeed changes). If we don't solve directly there, would be better to postpone these fixes after that pr gets merged.

alright, got it. thanks @altafan

PR #159, opened 2026-04-29
Auto-Settle

sekulicd:
## Summary

Adds auto-settle support and tests. The scheduler now computes the next settlement from spendable VTXOs plus confirmed boarding UTXOs, schedules at ExpiresAt - 2*SessionDuration, reacts to VTXO/UTXO events, and cancels safely on lock/stop. The e2e config increases VTXO expiry to 60s so WhenNextSettlement() can be observed deterministically before the timer fires.

@altafan please review.

## Summary by CodeRabbit

* **New Features**
  * Opt-in automatic settlement for offchain funds with a visible next-scheduled time (WhenNextSettlement) and a client option to enable it.

* **Behavior**
  * Scheduler starts after unlock/sync, listens for wallet events, schedules at expiry − 2×SessionDuration, fires immediately for expired funds, reschedules earlier candidates, cancels on lock/shutdown, and retries failures with exponential backoff.

* **Tests**
  * Unit and end-to-end tests covering scheduling, firing, reschedule/cancellation semantics, retries/backoff, and timing; test timing infra adjusted for longer expiries.


ghost (changes requested):
## Arkana Code Review — Auto-Settle (go-sdk#159)

Overall: well-structured feature with careful concurrency design. The `claim()` pattern, the dual cancel in Lock/Stop, and the hook-based test architecture are all solid. However, there is one protocol-relevant gap and a few items worth discussing before merge.

---

### 🔴 PROTOCOL-CRITICAL — No retry on settle failure

**`auto_settle.go:236-268`** — When the timer fires, `claim()` clears `nextSettleAt` and `settleTimer` *before* calling `hooks.settle(ctx)`. If `settle()` fails (network timeout, server round not starting, etc.), the scheduler has no pending schedule and no mechanism to retry.

Meanwhile the original VTXOs continue ticking toward expiry. No new `VtxosAdded` event will arrive to re-arm the timer (the settle didn't produce new VTXOs), so the user silently drifts toward on-chain expiry and the expensive unilateral exit path.

**Suggestion:** After a failed settle, re-scan current funds and reschedule:
```go
if err := hooks.settle(ctx); err != nil {
    log.WithError(err).Warn("auto-settle: scheduled settle failed")
    // Re-arm from current funds so we don't silently let VTXOs expire.
    a.scheduleFromCurrentFunds(ctx, cfg, hooks)
    return
}
```
Or at minimum, re-insert the *same* fire time with a small backoff so a transient failure doesn't abandon the schedule permanently. This is protocol-critical: undetected settlement failure → VTXO expiry → forced unilateral exit → higher fees and worse UX.

---

### 🟡 MEDIUM — `scheduleFromCurrentFunds` immediate settle doesn't reschedule remaining VTXOs

**`auto_settle.go:187-194`** — When at least one VTXO has already expired, the method calls `settle()` immediately and returns. It does not schedule a timer for the *remaining* non-expired VTXOs. The assumption is that settlement produces `VtxosAdded` events that the loop handles — but if the immediate settle only renews the expired VTXO (e.g., server only takes the expired one into the round), the other VTXOs' expiries are never scheduled until the next event. Worth a comment or a fallthrough to `scheduleSettle` for the rest.

---

### 🟡 MEDIUM — VTXO expiry config change affects all e2e tests

**`test/docker/docker-compose.yml:81`** — `ARKD_VTXO_TREE_EXPIRY` changed from `20` to `60`. This is a shared test infra config. All existing e2e tests that previously relied on the 20s window (e.g., tests waiting for expiry-related behavior) will now take 3× longer or behave differently. Consider isolating the auto-settle test's config or using a per-test override rather than changing the global default.

---

### 🟡 MEDIUM — Breaking interface addition

**`sdk.go:81-84`** — `WhenNextSettlement() time.Time` added to the `ArkClient` interface. Any external code implementing this interface from scratch (not via embedding a concrete `*arkClient`) will fail to compile. Fulmine embeds the interface and delegates, so it's fine today, but this is a public API break. Consider whether this belongs on the interface or could be a standalone function / separate interface that callers opt into.

---

### 🟢 LOW — `boardingUtxoExpiry` fallback with zero `SpendableAt`

**`auto_settle.go:341-344`** — When `SpendableAt` is zero, the fallback computes `CreatedAt.Add(delay)`. If `CreatedAt` is also zero (which `IsConfirmed()` likely prevents), this produces a time near year 0001 — filtered by `nextSettlementExpiry` but still surprising. A defensive `if utxo.CreatedAt.IsZero() { return ..., false }` would make the invariant explicit.

---

### 🟢 LOW — Non-atomic snapshot in `scheduleFromCurrentFunds`

**`auto_settle.go:171-179`** — `ListSpendableVtxos` doesn't hold `dbMu`; `listUnspentBoardingUtxos` does. The two snapshots are not atomic, so a concurrent `refreshDb` could update one between the calls. For the scheduler's purpose (finding earliest expiry) this is benign — any new event triggers rescheduling — but worth a comment documenting the intentional non-atomicity.

---

### ✅ What's good

- **Concurrency model:** The `claim()` / `settleTimer` identity check prevents double-fire cleanly. The explicit `cancelPendingSettle()` in both `Lock()` and `Stop()` handles the timer-fire race correctly.
- **Lock ordering:** `autoSettlement.mu` is released before `settle()` runs, avoiding deadlock with `dbMu` or event publishing.
- **Test coverage:** Unit tests cover scheduling, rescheduling, cancellation, locked-wallet, and canceled-context paths. The e2e test validates the full loop including event-driven rescheduling.
- **Delegate-mode guard:** Clean forward-looking wiring for delegate addresses.

---

⚠️ **This PR touches automatic VTXO settlement — protocol-critical functionality.** The retry-on-failure issue (🔴) should be addressed before merge. Requesting human review per protocol.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (changes requested):
## Arkana Follow-Up Review — New Commits (go-sdk#159)

Reviewed 3 new commits since my [previous review](https://github.com/arkade-os/go-sdk/pull/159#pullrequestreview-4197823012):
- `84b6b32` — merge from main
- `ed437fc` — formatting fix in `auto_settle_test.go` (line-wrapping `require.*` calls)
- `3f61a4f` — `batch_session_test.go`: bump `generateBlocks(41)` → `generateBlocks(81)` to accommodate the `ARKD_VTXO_TREE_EXPIRY` change from 20→60

### New changes assessment

The test formatting is fine. The `batch_session_test.go` fix (`generateBlocks(81)`) confirms the collateral damage I flagged in my previous review — the global `ARKD_VTXO_TREE_EXPIRY` change required patching an unrelated test. Worth checking if other e2e tests (exit, history, asset) need similar adjustments, or if they already passed CI.

### Previous findings — still open

**🔴 PROTOCOL-CRITICAL — No retry on settle failure** (`auto_settle.go:236-268`)
Not addressed. `auto_settle.go` was not modified. If `hooks.settle()` fails, the scheduler silently abandons the VTXO, which drifts toward unilateral exit. This must be fixed before merge.

**🟡 `scheduleFromCurrentFunds` immediate settle doesn't reschedule remaining VTXOs** (`auto_settle.go:187-194`)
Not addressed.

**🟡 Breaking `ArkClient` interface addition** (`sdk.go:81-84`)
Not addressed.

Maintaining request-changes until at minimum the 🔴 retry-on-failure issue is resolved.

⚠️ Protocol-critical functionality — human review required.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (changes requested):
## Arkana Follow-Up Review — Commit d47e9eaf (go-sdk#159)

Reviewed 1 new commit since my [previous review](https://github.com/arkade-os/go-sdk/pull/159#pullrequestreview-4199389838):
- `d47e9eaf` — `hd_wallet_test.go`: collapse separate `wantOnchainSpendable` + `wantLockedOnchain` assertions into a single `wantOnchainTotal` check.

### New commit assessment

**`test/e2e/hd_wallet_test.go` fix** — ✅ Looks correct. The spendable-vs-locked split is timing-dependent (depends on how many blocks have been confirmed when the assertion runs), so asserting the total (`SpendableAmount + LockedAmount == 129_000`) is more robust and still validates the correct aggregate balance. This is another expected side-effect of the `ARKD_VTXO_TREE_EXPIRY` 20→60 change.

### Previous findings — still open

All three findings from my earlier reviews remain unaddressed. `auto_settle.go` was not modified in any commit since my initial review.

**🔴 PROTOCOL-CRITICAL — No retry on settle failure** (`auto_settle.go:236-268`)
If `hooks.settle()` fails, the timer callback exits without rescheduling. The VTXO continues expiring toward unilateral exit, and the scheduler never fires again unless a *new* VtxosAdded event arrives. This is a fund-safety issue — a transient network error during a single Settle call silently abandons the VTXO to the forfeit path.

**🟡 `scheduleFromCurrentFunds` immediate settle doesn't reschedule remaining VTXOs** (`auto_settle.go:187-194`)
When at least one VTXO is already past expiry, `scheduleFromCurrentFunds` calls `hooks.settle()` and returns without scheduling the next-earliest unexpired VTXO. Remaining VTXOs are unprotected until a new event re-arms the scheduler.

**🟡 Breaking `ArkClient` interface addition** (`sdk.go:81-84`)
Adding `WhenNextSettlement()` to the `ArkClient` interface is a breaking change for any downstream implementation. Consider a separate interface (e.g. `AutoSettleClient`) or accept this as a major version bump.

Maintaining request-changes until at minimum the 🔴 retry-on-failure issue is resolved.

⚠️ Protocol-critical functionality — human review required before merge.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (commented):
## Arkana Follow-Up Review — Commit f4c21d05 (go-sdk#159)

Reviewed 1 new commit since my [previous review](https://github.com/arkade-os/go-sdk/pull/159#pullrequestreview-4199389838):
- `f4c21d05` — **auto settlement with backoff retry**

### 🔴 → ✅ Retry on settle failure — RESOLVED

This was my primary blocking finding. The new commit addresses it correctly:

- `settleWithRetry()` (`auto_settle.go:293-300`) wraps every `hooks.settle()` call. On failure, it schedules a retry via `scheduleSettleRetry()`. On success, it resets the backoff.
- `nextSettleRetryBackoffLocked()` (`auto_settle.go:314-337`) implements exponential backoff (5s initial, doubling to 1m cap). Clean and correct.
- `scheduleSettleAt()` properly separates event-driven schedules (`resetRetry=true`) from retry schedules (`resetRetry=false`), so a retry backoff isn't accidentally cleared by the retry's own reschedule.
- `cancelPendingSettle()` (`auto_settle.go:349`) correctly resets `retryBackoff` to zero on Lock/Stop, ensuring a fresh start on next Unlock.
- The skip-if-earlier guard in `scheduleSettleAt` correctly interacts with retries: if a new event produces a later fireAt than a pending retry, the retry takes priority (guard returns early, backoff preserved). If the event is earlier, it replaces the retry and resets backoff. Both behaviors are correct.
- Test coverage (`auto_settle_test.go:109-147`) validates the retry path with channel synchronization rather than sleeps. Solid.

### 🟡 → ✅ `scheduleFromCurrentFunds` immediate settle — RESOLVED

Now uses `settleWithRetry()` (`auto_settle.go:196`). My previous concern about remaining unexpired VTXOs was overstated — `Settle()` settles *all* spendable VTXOs, so after success the event loop will pick up new VTXOs from the round via `VtxosAdded` and reschedule as needed. No gap.

### 🟡 Breaking `ArkClient` interface addition — still open

`sdk.go` was not modified. Adding `WhenNextSettlement()` to the `ArkClient` interface remains a breaking change for downstream implementors. Not blocking for fund safety, but worth addressing before merge if there are external consumers.

### New code assessment

The retry implementation is well-structured. No new concerns:
- No unbounded goroutine leaks (retries go through `time.AfterFunc`, same as normal scheduling)
- Context cancellation is checked at every entry point (`scheduleSettleAt:215`, `scheduleSettleRetry:303`)
- Mutex discipline is correct — `nextSettleRetryBackoffLocked()` is called under lock, `resetSettleRetryBackoff()` acquires its own lock
- The `claim()` closure inside `AfterFunc` prevents stale timers from firing after replacement

**Removing request-changes.** The fund-safety concern is resolved. The interface change is a minor API design issue, not a blocker.

⚠️ Protocol-critical functionality — human review still required before merge.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (approved):
Approving — the 🔴 retry-on-failure issue is resolved. See my detailed comment above.

Remaining minor item: `WhenNextSettlement()` added to `ArkClient` interface is a breaking change for external implementors. Consider a separate interface or accept as intentional.

⚠️ Protocol-critical — human approval still required.

ghost (approved):
## Arkana Follow-Up Review — Commit 20eb3732 (go-sdk#159)

Reviewed 1 new commit since my [previous review](https://github.com/arkade-os/go-sdk/pull/159#pullrequestreview-4199464150):
- `20eb3732` — **retry if initial fetch fails** — adds snapshot retry with exponential backoff

### New commit assessment — ✅ Clean

This addresses CodeRabbit's finding about `scheduleFromCurrentFunds` silently giving up when `ListSpendableVtxos` or `listUnspentBoardingUtxos` fails on startup. Good fix — without it, a transient RPC failure during Unlock would leave the scheduler blind to all pre-existing VTXOs until a new event arrived.

**What changed:**
1. Three new hooks (`getConfigData`, `listSpendableVtxos`, `listUnspentBoardingUtxos`) added to `autoSettleHooks` for testability.
2. `resolveAutoSettleHooks()` now does field-by-field merging instead of all-or-nothing — partial hook overrides in tests work correctly.
3. `scheduleFromCurrentFunds` returns `bool` — `true` on successful read (even if no funds), `false` on read failure.
4. Snapshot retry timer integrated into the event loop `select`, so VTXO/UTXO events continue to be processed while waiting to retry. No goroutine leak, no blocking.
5. `nextAutoSettleRetryBackoff` extracted as a pure function, shared between settle retry and snapshot retry. DRY and testable.

**Verified correct:**
- **No race conditions**: `snapshotRetryTimer`, `snapshotRetryCh`, `snapshotRetryBackoff` are local to `autoSettleLoop` — single goroutine access only.
- **Timer cleanup**: `stopSnapshotRetry()` is deferred and properly drains the channel on `Stop()`.
- **Retry/event interaction**: Events arriving during snapshot retry are handled independently by their own `select` cases. The `scheduleSettleAt` skip-if-earlier guard prevents redundant rescheduling when the snapshot retry eventually succeeds with a later expiry than an already-scheduled event. Both directions are correct.
- **`scheduleSnapshotRetry` guard** (`snapshotRetryTimer != nil || ctx.Err() != nil`): prevents double-arming. After the timer fires, `snapshotRetryTimer` is nil'd before calling `scheduleFromCurrentFunds`, so re-arming on failure works correctly.
- **Backoff continuity on repeated failure**: On retry fire, `snapshotRetryBackoff` is preserved (only reset to 0 on success), so `scheduleSnapshotRetry` correctly doubles. Verified the path: fire → nil timer → fail → `scheduleSnapshotRetry()` → `nextAutoSettleRetryBackoff(previous, ...)` → doubled.
- **`getConfigData` failure still aborts**: Config fetch failure remains fatal (no retry), which is correct — config is static and a failure here likely means the client is broken, not a transient issue.

**Test** (`TestAutoSettleLoopRetriesInitialSnapshotWithoutBlockingEvents`): Well-structured. Verifies (a) events are processed while snapshot retry is pending, (b) retry fires and succeeds, (c) retry stops after success (no infinite loop). Uses channel synchronization instead of sleeps for the critical paths.

### All previous findings — resolved

No open issues remain. Maintaining approval.

⚠️ Protocol-critical functionality — human review still required before merge.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

client_opts.go
```diff
@@ -77,6 +77,19 @@ func WithWallet(walletSvc wallet.WalletService) ClientOption {
...
+// spendable VTXO expiry. Off by default; explicit opt-in required.
+//
+// The feature is a no-op when delegate addresses are in use (a warning is
+// logged when the loop starts).
+func WithAutoSettle() ClientOption {
+	return func(o *clientOptions) error {
+		o.autoSettle = true
+		return nil
+	}
+}
```
altafan, line 90: since we added the delegate support already, let's already add the check that makes the 2 optiosn mutually exclusive, this way we won't forget to add it later in the upcming option

ghost (approved):
## Arkana Follow-Up Review — Commits 83199fc8, 4bdce592 (go-sdk#159)

Reviewed 2 new commits since my [previous review](https://github.com/arkade-os/go-sdk/pull/159#pullrequestreview-4199464150):
- `83199fc8` — **add logging** — adds `log.Info`/`log.WithFields` on schedule and successful settlement
- `4bdce592` — **pr review** — removes internal `REQ-*` comment markers, tightens `WithAutoSettle` doc, adds hard `applyClientOptions` validation rejecting `autoSettle + delegateMode`

### Assessment — ✅ LGTM

**Logging (`83199fc8`):** Good additions. `settle_at`, `delay`, and `retry` fields give operators enough to debug scheduling without leaking sensitive data. No hot-path concern — these fire at most once per VTXO expiry cycle.

**Delegate-mode guard (`4bdce592`):** Previously the delegate guard was a runtime no-op (log + return). Now it's a hard construction-time error in `applyClientOptions`, which is strictly better — fail fast instead of silently degrading. The runtime guard in `autoSettleLoop` is still present as defense-in-depth, which is correct.

**Comment cleanup:** `REQ-*` references removed. No functional change.

No new protocol-critical concerns. Previous review findings still apply; this incremental delta is clean.

---
*⚠️ Reminder: this PR adds auto-settlement (protocol-critical — triggers `Settle()` automatically). Human approval still required before merge.*

client_opts.go
sekulicd: Done

ghost (approved):
## Arkana Follow-Up Review — Commit 30730ea (go-sdk#159)

Reviewed 1 new commit since my [previous review](https://github.com/arkade-os/go-sdk/pull/159#pullrequestreview-4211949940):
- `30730ea` — **update comments** — improves doc comments on `autoSettleLoop`, `scheduleFromCurrentFunds`, snapshot retry helpers, and nil-channel behavior

### Assessment — ✅ No-op (comments only)

Single file changed (`auto_settle.go`), 27 insertions / 16 deletions — all within Go doc comments and inline comments. Zero functional code changes. Verified via `git diff --stat`.

The updated comments are more precise:
- Lifecycle step 1 now correctly notes the defense-in-depth nature of the delegate guard (option validation is the primary gate)
- Step 4 documents the snapshot retry path added in `20eb3732`
- Nil-channel comment clarifies `select` semantics (disabled case vs. blocking forever)
- `scheduleFromCurrentFunds` godoc now explains the bool return value contract
- `stopSnapshotRetry` and `scheduleSnapshotRetry` closures now have proper header comments

No protocol-critical concerns. Maintaining approval.

⚠️ Reminder: this PR adds auto-settlement (protocol-critical — triggers `Settle()` automatically). Human approval still required before merge.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (approved):
## Arkana Follow-Up Review — Commit e9ecb891 (go-sdk#159)

Reviewed 1 new commit since my [previous review](https://github.com/arkade-os/go-sdk/pull/159#pullrequestreview-4211949940):
- `e9ecb891` — **comments update** — rewrites doc comments for clarity; simplifies `stopSnapshotRetry`

### Assessment — ✅ Clean

Single file changed (`auto_settle.go`), +19 / -24 — almost entirely comment rewording.

**One functional delta:** `stopSnapshotRetry` no longer drains the timer channel after `Stop()`. This is correct — the closure only runs in the `defer` when `autoSettleLoop` exits, at which point `snapshotRetryCh` has no readers, so a buffered tick is harmless and will be GC'd with the timer.

No protocol-critical concerns. Maintaining approval.

⚠️ Reminder: this PR adds auto-settlement (protocol-critical — triggers `Settle()` automatically). Human approval still required before merge.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

altafan:
Contains improvements that can def be added after the replacement of this (#169) has been merged

PR #162, opened 2026-04-29
fix(hdwallet): sort derived keys by numeric derivation index

### LINKED ISSUE #161: ListKeys returns keys in wrong order for wallets with 10 or more addresses
When a user has generated 10 or more addresses, ListKeys returns them in the wrong order. This is because the code sorts the keys as text (like words in a dictionary) rather than as numbers. In text sorting, "10" comes before "2"... the same way "cat" comes before "dog" because it only looks at the first character.

https://github.com/arkade-os/go-sdk/blob/6ce96d161238cf7fd284b35662f8a27fd2909032/wallet/hdwallet/wallet.go#L304-L306

This doesn't affect users with fewer than 10 addresses because single digit numbers sort the same way whether you treat them as text or numbers.

Dunsin-cyber:
Fixes #161  where `ListKeys` sorted key derivation paths as strings instead of numbers. This caused keys at index 10 and above to appear before index 2, 3 etc. in the returned list, which could cause wallet recovery to miss funds when scanning derived addresses. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Keys are now returned in derivation order rather than alphabetical order.
  * Enhanced error handling for invalid key data during sorting operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/go-sdk/pull/162)


ghost (changes requested):
## Review: fix(hdwallet): sort derived keys by numeric derivation index

Good catch on the lexicographic vs numeric sort — this is a real bug that would cause wallet recovery to scan addresses in wrong order once ≥10 keys are derived. The test update to 12 keys properly exercises the boundary. However, the fix has a latent crash bug that needs addressing.

---

### 🔴 Bug: panic on `parseDerivationIndex` error — `wallet.go:305-306`

```go
pi, _ := parseDerivationIndex(keys[i].Id)
pj, _ := parseDerivationIndex(keys[j].Id)
return pi[1] < pj[1]
```

If `parseDerivationIndex` returns an error, it returns `nil` for the slice. Accessing `pi[1]` on a `nil` slice is an **index-out-of-range panic**. The error is silently discarded with `_`.

Yes, in normal flow key IDs are always generated by `derivationPath(index)` → `"m/0/{N}"`, so `parseDerivationIndex` will always succeed. But this is a wallet — a panic in `ListKeys` crashes the entire application. Defensive coding is non-negotiable here.

**Recommended fix:** Pre-compute numeric indices before sorting so errors can be surfaced properly. This also avoids redundant parsing during sort comparisons:

```go
type indexedKey struct {
    ref   wallet.KeyRef
    index uint32
}

indexed := make([]indexedKey, 0, len(keys))
for _, k := range keys {
    path, err := parseDerivationIndex(k.Id)
    if err != nil {
        return nil, fmt.Errorf("invalid key id %q: %w", k.Id, err)
    }
    indexed = append(indexed, indexedKey{ref: k, index: path[1]})
}

sort.SliceStable(indexed, func(i, j int) bool {
    return indexed[i].index < indexed[j].index
})

result := make([]wallet.KeyRef, len(indexed))
for i, ik := range indexed {
    result[i] = ik.ref
}
return result, nil
```

Alternatively, a simpler (minimal-diff) approach if you want to keep the current structure — just guard the nil:

```go
sort.SliceStable(keys, func(i, j int) bool {
    pi, ei := parseDerivationIndex(keys[i].Id)
    pj, ej := parseDerivationIndex(keys[j].Id)
    if ei != nil || ej != nil {
        return keys[i].Id < keys[j].Id // fallback to string sort
    }
    return pi[1] < pj[1]
})
```

But I'd prefer the pre-compute approach — it surfaces errors to the caller and avoids O(n log n) repeated parsing.

---

### ✅ Test changes look good

Expanding from 3 to 12 keys properly covers the `10+` boundary where string sort diverges from numeric sort. The assertion that `allocated[i] == keys[i].Id` confirms stable ordering matches allocation order.

---

### Cross-repo impact: None

Checked all consumers of `ListKeys`:
- `client.go:388` — only checks `len(keys) > 0`, order-independent
- `funding.go:194` — iterates all keys with index filtering, order-independent
- No downstream repos (arkd, ts-sdk, etc.) consume this Go function directly

**Summary:** The fix is directionally correct and addresses a real bug, but the silent error discard creates a panic vector. Please handle the error path, then this is good to merge.

ghost (approved):
## Re-review: error handling fix

Previous concern (panic on `parseDerivationIndex` error) is resolved. The closure-captured `sortErr` approach is correct:

- ✅ No nil-slice access — `pi[1]`/`pj[1]` gated behind error checks
- ✅ Error surfaced to caller — `sortErr` checked post-sort
- ✅ No race — `sort.SliceStable` is single-threaded
- ✅ Test covers the 10+ key boundary where lex vs numeric sort diverges

Minor style nit (non-blocking): pre-computing indices before sort would be cleaner (O(n) parsing, no sort-contract violation on error), but correctness is satisfied as-is. Ship it.

🤖 `arkana`

altafan (approved):
@Dunsin-cyber nice catch! thank you for the fix 🙏

Dunsin-cyber:
> @Dunsin-cyber nice catch! thank you for the fix 🙏

my pleasure 😊

Dunsin-cyber:
is this good to go? 
CC @bitcoin-coder-bob  @altafan @louisinger

bitcoin-coder-bob:
> is this good to go? CC @bitcoin-coder-bob @altafan @louisinger

LGTM

PR #167, opened 2026-05-05
Fix RBF replacement mapping outputs by index instead of script

### LINKED ISSUE #166: RBF bumpfee on boarding tx can cause INVALID_PSBT_INPUT on settle due to output index reordering
When a user funds a boarding address and then bumps the fee (bumpfee), the subsequent settle call fails with:                                                                             
   
 ```
 INVALID_PSBT_INPUT (5): invalid witness utxo script: got 512057e233fb803c0db0fe3d3aa0d6ea9b53028fb34e245980991d0971d48c39da21 expected                                                    
  512010a0fefd8dd80db429bf780471f71fcb6959cfe27c60d724a40971aa50c71480                                                                                                                      
   
```
  Root Cause                                                                                                                                                                                
                  
  In client.go, the RBF replacement handler maps old UTXO outpoints to replacement tx outpoints by output index:                                                                            
                  
```
  for outputIndex := range tx.TxOut {                                                                                                                                                       
      replacedUtxo := clientTypes.Outpoint{Txid: replacedTxid, VOut: uint32(outputIndex)}                                                                                                   
      // ...                                                                                                                                                                                
      utxoStore.ReplaceUtxo(ctx, replacedUtxo, clientTypes.Outpoint{                                                                                                                        
          Txid: replacementTxid,                                                                                                                                                            
          VOut: uint32(outputIndex), // assumes same index
      })                                                                                                                                                                                    
  }               
```
                                                                                                                                                                                            
  Bitcoin Core's bumpfee can reorder outputs. If the original tx has [change@0, boarding@1] and the replacement has [boarding@0, change@1], the handler maps (oldTxid, 1) → (newTxid, 1) —  
  but newTxid:1 is now the change output, not the boarding output.
                                                                                                                                                                                            
  When settle builds the PSBT, it references (newTxid, 1) with the boarding script. arkd looks up (newTxid, 1) on-chain, finds the change script, and rejects with INVALID_PSBT_INPUT. 

Check TXs in prod:
https://mempool.space/tx/ec12bf32ea88a28de56568d73edbec696ccc0e5f10f3694abd3895d751a22ee6
https://mempool.space/tx/1f32462eacf5896b9e318424d9706e46366257b32a5e7aaf2e634b12da9a3f4e

UTXO table:
sqlite> select txid,vout from utxo;
1adca93434544147577b5f4cd770125e0f0967560b0502e4e504822e813c00ce|1
1f32462eacf5896b9e318424d9706e46366257b32a5e7aaf2e634b12da9a3f4e|0
1f32462eacf5896b9e318424d9706e46366257b32a5e7aaf2e634b12da9a3f4e|1
dc219796e670e1598b946cc91a5f986c8b5706c4fb61d5483a9947c60e94a59f|1

### PR DESCRIPTION
**Pull request by sekulicd**
Close #166 

## Summary      

  - **RBF output reordering fix**: When handling boarding tx replacements (bumpfee), match stored UTXOs to replacement tx outputs by **script** instead of assuming output indices are      
  preserved. Bitcoin Core's `bumpfee` can reorder outputs, causing the stored UTXO to point to the wrong output (e.g. change instead of boarding).
  - **HD discovery guard**: Skip `discoverHDWalletKeys` for non-HD wallets (e.g. single-key). The single-key wallet's `GetKey` ignores the keyID parameter and always returns the same key, 
  which causes the gap-limit discovery loop to run forever (every index returns the same "used" key).                                                                                       
   
  ## Root Cause (RBF)                                                                                                                                                                       
                  
  Original tx:     [change@vout0, boarding@vout1]                                                                                                                                           
  Replacement tx:  [boarding@vout0, change@vout1]  ← reordered
                                                                                                                                                                                            
  Old code: (oldTxid, 1) → (newTxid, 1)  ← now points to change output                                                                                                                      
  Fix:      (oldTxid, 1) → (newTxid, 0)  ← matched by script
                                                                                                                                                                                            
  arkd validates the PSBT witness UTXO script against what's on-chain. Wrong index → wrong script → `INVALID_PSBT_INPUT`.
                                                                                                                                                                                            
  ## Root Cause (HD discovery)
                                                                                                                                                                                            
  `discoverHDWalletKeys` calls `GetKey("m/0/0")`, `GetKey("m/0/1")`, ... checking for VTXO activity. Single-key wallet ignores the path and always returns the same key. If that key has    
  VTXOs, every index appears "used", `consecutiveUnused` never reaches gapLimit, and the loop never terminates — spamming `GetVtxos` to arkd indefinitely.

@altafan please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Replace-By-Fee (RBF) handling to match UTXOs by script rather than output index, preventing incorrect replacements when outputs are reordered and ensuring settle succeeds after fee-bumps
  * Fixed wallet discovery to correctly skip non-HD wallets and avoid unintended gap-limit scanning

* **Tests**
  * Added unit and end-to-end tests covering RBF replacements, reordered outputs, and settle behavior after bump-fee operations

### CODE REVIEW

File: client.go [hunk 1]
```diff
@@ -1031,28 +1031,41 @@ func (a *arkClient) listenForOnchainTxs(ctx context.Context) {
 
 					utxoStore := a.store.UtxoStore()
 
-					for outputIndex := range tx.TxOut {
-						replacedUtxo := clientTypes.Outpoint{
+					// Collect all stored UTXOs belonging to the replaced
```
tiero: doesn't sounds the right approach? Why we don't simply follow the replacement tx via looking at the mempool directly to know what the latest?

sekulicd: We already follow the replacement tx from the mempool — update.Replacements gives us the new txid and we fetch the full tx via GetTxHex. The bug isn't about finding the replacement tx,
   it's about mapping the output index. The old code did (oldTxid, vout=1) → (newTxid, vout=1), assuming indices are preserved. But bumpfee can reorder outputs, so the boarding output at 
  vout 1 in the original might be at vout 0 in the replacement. Matching by pkScript is how we find the correct output in the new tx.

tiero: ok comment seems off, why we do need probing if we actually store the script to retrieve it cleanly?

Kukks: This is weird. We shouldnt update a record. Rather mark it as replaced and add a new record with the new outpoint?

client.go
```diff
@@ -1031,28 +1031,41 @@ func (a *arkClient) listenForOnchainTxs(ctx context.Context) {
...
-								VOut: uint32(outputIndex),
-							})
-							a.dbMu.Unlock()
-							if err != nil {
-								log.WithError(err).Error("failed to replace boarding utxo")
-								continue
-							}
+						if err == nil {
+							storedUtxos = append(storedUtxos, utxos...)
+						}
```
altafan: let's make a single call to db with the list of all outpoints instead of a call per outpoint

client.go
```diff
@@ -2179,3 +2192,44 @@ func (a *arkClient) lookupTrackedAddressByScript(script string) (trackedAddressI
 	info, ok := a.onchainRegistry.addressByScript[script]
 	return info, ok
 }
+
+// utxoReplacement represents a mapping from an old UTXO outpoint to its
+// replacement outpoint after an RBF (Replace-By-Fee) transaction.
+type utxoReplacement struct {
+	From clientTypes.Outpoint
+	To   clientTypes.Outpoint
+}
```
altafan, line 2195:
```suggestion
type utxoReplacement struct {
	from clientTypes.Outpoint
	to   clientTypes.Outpoint
}
```

ghost (commented):
## Arkana Code Review — `fix/rbf-output-index-mapping`

Good fix for a real production bug. The script-based matching for RBF output reordering is the correct approach, and the HD discovery guard prevents an infinite loop. Test coverage is solid. A few findings:

---

### 🟡 Bug: Outpoint probe uses replacement tx output count (Medium)

**`client.go:1037-1044`** — The code iterates `range tx.TxOut` (the *replacement* tx) to build probe outpoints for the *replaced* txid. If the replacement tx has **fewer** outputs than the original (e.g., change output absorbed into fees), stored UTXOs at higher indices in the old tx will never be probed and become orphaned in the store.

```go
// Current: probes based on replacement tx output count
for i := range tx.TxOut {  // tx = replacement transaction
    outpoint := clientTypes.Outpoint{
        Txid: replacedTxid,  // but using old txid
        VOut: uint32(i),
    }
```

**Fix:** Either fetch the replaced tx to determine its output count, or use a broader approach — query all UTXOs by txid prefix if the store supports it. Alternatively, since `bumpfee` in practice preserves or reduces output count, this may be acceptable if documented as a known limitation.

---

### 🟡 Duplicate script edge case (Low-Medium)

**`client.go:2218-2235`** — `matchReplacementOutputs` matches by script alone. If a single tx has multiple outputs with the **same script** (uncommon but valid — e.g., two boarding deposits to the same address in one tx), the pairing is arbitrary based on iteration order. The `matched` map prevents double-assignment of new outputs, but doesn't guarantee correct pairing.

Since amounts can change in RBF, matching by `(script, amount)` isn't reliable either. This is probably acceptable for now but worth a comment noting the limitation.

---

### 🟢 Dead code: `DeleteUtxos` (Informational)

`DeleteUtxos` is added to the `UtxoStore` interface (`types/interfaces.go:37`) and both store implementations, but is **never called** anywhere in this PR. If this is prep for a follow-up PR, consider noting that in the PR description. Otherwise it's dead code expanding the interface contract.

---

### 🟢 `Store()` exposed on public interface (Informational)

**`sdk.go:24`** — Adding `Store() types.Store` to the public `ArkClient` interface exposes internal store state to all consumers. If this is only needed for testing, consider a separate test-only interface or unexported accessor.

---

### ✅ What looks good

- Script-based matching logic in `matchReplacementOutputs` is correct and well-tested
- The `matched` map prevents double-mapping — good
- HD discovery guard (`discover.go:34`) is a clean, minimal fix for the infinite loop
- Unit tests cover: same order, reordered, missing output, multiple UTXOs
- E2E test (`test/e2e/rbf_test.go`) validates the full flow with real `bumpfee`
- Error handling on `GetUtxos` failure properly `continue`s the loop

---

**Verdict:** The core fix is sound. The outpoint probe issue (finding #1) is the most concerning — in a scenario where `bumpfee` removes an output entirely, a stored UTXO could be orphaned. Please confirm whether this case is possible in practice or add a safeguard.

⚠️ This touches UTXO tracking which is protocol-adjacent. Requesting human review.

ghost (commented):
## Arkana Code Review — `fix/rbf-output-index-mapping`

Good fix for a real production bug. The script-based matching for RBF output reordering is the correct approach, and the HD discovery guard prevents an infinite loop. Test coverage is solid. A few findings:

---

### 🟡 Bug: Outpoint probe uses replacement tx output count (Medium)

**`client.go:1037-1044`** — The code iterates `range tx.TxOut` (the *replacement* tx) to build probe outpoints for the *replaced* txid. If the replacement tx has **fewer** outputs than the original (e.g., change output absorbed into fees), stored UTXOs at higher indices in the old tx will never be probed and become orphaned in the store.

```go
// Current: probes based on replacement tx output count
for i := range tx.TxOut {  // tx = replacement transaction
    outpoint := clientTypes.Outpoint{
        Txid: replacedTxid,  // but using old txid
        VOut: uint32(i),
    }
```

**Fix:** Either fetch the replaced tx to determine its output count, or use a broader approach — query all UTXOs by txid prefix if the store supports it. Alternatively, since `bumpfee` in practice preserves or reduces output count, this may be acceptable if documented as a known limitation.

---

### 🟡 Duplicate script edge case (Low-Medium)

**`client.go:2218-2235`** — `matchReplacementOutputs` matches by script alone. If a single tx has multiple outputs with the **same script** (uncommon but valid — e.g., two boarding deposits to the same address in one tx), the pairing is arbitrary based on iteration order. The `matched` map prevents double-assignment of new outputs, but doesn't guarantee correct pairing.

Since amounts can change in RBF, matching by `(script, amount)` isn't reliable either. This is probably acceptable for now but worth a comment noting the limitation.

---

### 🟢 Dead code: `DeleteUtxos` (Informational)

`DeleteUtxos` is added to the `UtxoStore` interface (`types/interfaces.go:37`) and both store implementations, but is **never called** anywhere in this PR. If this is prep for a follow-up PR, consider noting that in the PR description. Otherwise it's dead code expanding the interface contract.

---

### 🟢 `Store()` exposed on public interface (Informational)

**`sdk.go:24`** — Adding `Store() types.Store` to the public `ArkClient` interface exposes internal store state to all consumers. If this is only needed for testing, consider a separate test-only interface or unexported accessor.

---

### ✅ What looks good

- Script-based matching logic in `matchReplacementOutputs` is correct and well-tested
- The `matched` map prevents double-mapping — good
- HD discovery guard (`discover.go:34`) is a clean, minimal fix for the infinite loop
- Unit tests cover: same order, reordered, missing output, multiple UTXOs
- E2E test (`test/e2e/rbf_test.go`) validates the full flow with real `bumpfee`
- Error handling on `GetUtxos` failure properly `continue`s the loop

---

**Verdict:** The core fix is sound. The outpoint probe issue (finding #1) is the most concerning — in a scenario where `bumpfee` removes an output entirely, a stored UTXO could be orphaned. Please confirm whether this case is possible in practice or add a safeguard.

⚠️ This touches UTXO tracking which is protocol-adjacent. Requesting human review.

test/e2e/rbf_test.go
```diff
@@ -0,0 +1,99 @@
...
+		require.NoError(t, generateBlocks(1))
+	}
+
+	// Wait for the SDK to detect the confirmed UTXOs, then settle.
+	require.Eventually(t, func() bool {
+		_, err := client.Settle(ctx)
+		return err == nil
+	}, 60*time.Second, 2*time.Second, "settle should succeed after RBF bumpfee")
+
+	t.Log("settle succeeded after RBF bumpfee")
```
altafan: remove log

ghost (approved):
## Arkana Code Review — Follow-up (new commits)

Reviewed commits `a0b476b0` through `9c4f263b` (altafan + sekulicd post-review fixes).

### ✅ Primary finding resolved

My previous review flagged that the outpoint probe iterated `range tx.TxOut` (replacement tx) to build old-txid outpoints — meaning stored UTXOs at indices beyond the replacement tx output count would be silently orphaned. 

**Fixed correctly:** `client.go` now calls `utxoStore.GetUtxosByTxid(ctx, replacedTxid)` which fetches ALL stored UTXOs for the replaced txid directly from the store. No more index guessing. This is the right approach.

### ✅ New code looks clean

- `GetUtxosByTxid` implementations (KV: `badgerhold.Where("Txid").Eq(txid)`, SQL: `SELECT * FROM utxo WHERE txid = ?1`) — both correct and consistent.
- `DeleteUtxos` error handling improved: KV store now properly distinguishes `ErrNotFound` (skip) from real errors (return), matching the SQL store behavior.
- Test coverage for `GetUtxosByTxid` covers multi-output lookup and unknown-txid-returns-empty.

### Remaining notes from previous review (unchanged, acceptable)

- **Duplicate script edge case:** `matchReplacementOutputs` still matches by script alone. Two outputs with identical scripts in one tx get arbitrary pairing. Uncommon in practice; noted as known limitation.
- **`DeleteUtxos` unused in prod:** Only called in tests. Fine as interface completeness.
- **`Store()` on public interface:** Still exposed. Acceptable if needed by consumers.

**Verdict:** All protocol-relevant concerns addressed. Approving.

⚠️ This PR touches UTXO tracking and RBF replacement logic — **human review still required** before merge.

altafan (approved):

PR #173, opened 2026-05-12
Remove KV store

### LINKED ISSUE #142: Remove KV store
We currently maintain two storage backends (SQLite and Badger), which increases maintenance overhead. Tests also run against both, making them slower.

sekulicd:
This removes KV store so we dont need to maintain double repos.

Closes #142 

@altafan please review.

## Summary by CodeRabbit

* **Refactor**
  * App data and client config now always use the SQL (SQLite) backend; legacy KV/in-memory backends removed.
  * Storage initialization requires a non-empty data directory.
  * Centralized database lifecycle handling; individual sub-stores no longer expose separate close operations.

* **Tests**
  * Test helpers and suites updated to run using the SQL-backed store and temporary data directories.


ghost (changes requested):
## Code Review: Remove KV store (#173)

Good cleanup overall — removing a redundant storage backend is the right direction. But there are **breaking changes and a bug** that need to be addressed before this ships.

---

### 🔴 BUG: `LoadArkClient` with empty `datadir` will create a SQLite file in cwd

**`client.go:159-160`** — When `datadir` is empty, `dbConfig` is now `store.Config{}` (empty `BaseDir`). The new `NewStore()` unconditionally calls `filepath.Join("", sqliteDbFile)`, which resolves to just the filename in the current working directory.

Previously, empty datadir + `KVStore` type used in-memory BadgerDB. Now it silently writes a SQLite file to wherever the process happens to be running.

**Fix:** Either error on empty `BaseDir` in `NewStore`, or preserve the empty-datadir-returns-nil-stores behavior for `LoadArkClient`.

---

### 🔴 BREAKING: `NewArkClient("")` now errors

**`client.go:87-89`** — Added `if len(datadir) == 0 { return nil, errors.New("datadir must be specified") }`.

Previously, `NewArkClient("")` was valid and used in-memory stores. These downstream callers will break:
- `demos/asset/golang/main.go:128` — `arksdk.NewArkClient("")`
- `bancod/test/e2e/main_test.go:103` — `arksdk.NewArkClient("")`
- `bancod/test/e2e/utils_test.go:105` — `arksdk.NewArkClient("")`
- `asset-demos/asset/golang/main.go:128` — `arksdk.NewArkClient("")`

Needs coordinated PRs on those repos, or document the migration.

---

### 🟡 BREAKING: `types.KVStore` constant removed

**`types/types.go`** — `KVStore = "kv"` constant removed. Active consumers:
- `layerzero-usdt0-arkade-demo/test/e2e_flow_test.go:400`
- `introspector-review/test/tx_test.go:1310`

These compile against the published module. They'll break on the next `go get` update. PRs needed there too.

---

### 🟡 Dead code: `Config.AppDataStoreType` is now ignored

**`store/service.go:39`** — `NewStore` no longer reads `storeConfig.AppDataStoreType` at all. It unconditionally opens SQLite. The `AppDataStoreType` field in `Config` is dead code but still exists, which is misleading. Either remove it or add a guard that rejects unknown values.

---

### 🟢 No data migration path

Existing users with data in BadgerDB KV stores will lose access to their wallet data (UTXOs, VTXOs, transactions, contracts) when upgrading. This should be documented in a migration guide or changelog entry, even if the expectation is that users re-sync.

For a wallet SDK, silently losing VTXO state is dangerous — users could miss expiring VTXOs and lose funds.

---

### ✅ What looks good

- The deleted KV code (`store/kv/*`) is clean removal, no orphaned references
- Test updates correctly remove KV backends from `forEachContractBackend` and other table-driven tests
- `go.sum` cleanup removes ~163 lines of badger/grpc transitive deps — nice
- Comment updates (`KV store` → `SQLite store`) are accurate
- No protocol-critical logic changes — this is purely storage layer

---

**Verdict:** Request changes. Fix the `LoadArkClient` empty-datadir bug, coordinate the downstream breakage, and clean up the dead `AppDataStoreType` field.

ghost (approved):
## Re-review after new commits (4fe86f6, d583bef)

All issues from my previous review are addressed:

| Finding | Status |
|---------|--------|
| 🔴 `LoadArkClient` empty datadir → SQLite in cwd | ✅ Fixed — early validation at `client.go:155-158` |
| 🔴 Trim-before-check ordering in `NewArkClient` | ✅ Fixed — `TrimSpace` now precedes empty check (`client.go:87-88`) |
| 🟡 `AppDataStoreType` dead code | ✅ Fixed — `NewStore` validates it (`store/service.go:39-41`) |
| 🟡 DB handle leak on init failure | ✅ Fixed — `succeeded` pattern (`store/service.go:49-55,76`) |

Tests are solid — empty, whitespace-only, and non-creatable datadir cases covered for both `NewArkClient` and `LoadArkClient`.

**Remaining note (non-blocking):** Downstream repos using `NewArkClient("")` or `types.KVStore` will break on upgrade. Coordinate PRs or document the migration in a changelog.

LGTM. Still needs human approval per protocol.

store/service.go
```diff
@@ -37,75 +36,57 @@ type Config struct {
...
 
-	return &service{utxoStore, vtxoStore, txStore, assetStore, contractStore}, nil
+	succeeded = true
+	utxoStore := sqlstore.NewUtxoStore(db)
+	vtxoStore := sqlstore.NewVtxoStore(db)
+	txStore := sqlstore.NewTransactionStore(db)
+	assetStore := sqlstore.NewAssetStore(db)
+	contractStore := sqlstore.NewContractStore(db)
+
+	return &service{
```
altafan, line 85: now that we removed the kv store and all underlying stores (utxo, vtxo, tx, ...) share the same db, we can fix `service.Close` and make it close the connection to the db instead of delegating, which would repeat the same action 5 times.

altafan: no need to make any change to the interfaces, we just skip calling utxoStore.Close etc.

altafan (approved):

PR #175, opened 2026-05-14
Add paginated VTXO listing with filter support

sekulicd:
## Summary                                                                                                                                                                                
                  
  - Adds offset-based pagination to VTXO listing via `Page{PageNum, PageSize}` with `MaxPageSize=200`                                                                                       
  - Merges `ListVtxos` and `ListSpendableVtxos` into a single `ListVtxos(ctx, page, filter)` method
  - Introduces `VtxoFilter` enum: `All`, `Spendable`, `Spent`, `Recoverable`                                                                                                                
  - `Page{}` zero-value means "return all" — backward compatible for internal callers                                                                                                       
  - SQL pagination uses subquery on `vtxo` table (not the `asset_vtxo_vw` view) to correctly handle multi-asset VTXOs at page boundaries                                                    

                                                                                                                                                                                          
  ### Motivation                                                                                                                                                                            
                  
  Addresses the root cause behind [ArkLabsHQ/fulmine#405](https://github.com/ArkLabsHQ/fulmine/issues/405), where unbounded `GetVtxos` responses exceed gRPC's default 4 MiB receive limit  
  on wallets with large VTXO history (12k+ VTXOs observed in production). Raising the gRPC limit is a bandage — pagination is the proper fix.
                                                                                                                                                                                            
  ### API Changes (breaking)

  ```go
  // Before (two methods, no pagination):
  ListVtxos(ctx) (spendable, spent []Vtxo, err error)                                                                                                                                       
  ListSpendableVtxos(ctx) ([]Vtxo, error)                                                                                                                                                   
                                                                                                                                                                                            
  // After (single method with pagination + filter):                                                                                                                                        
  ListVtxos(ctx, page Page, filter VtxoFilter) ([]Vtxo, error)

Filter semantics
                                                                                                                                                                                            
  ┌───────────────────────┬─────────────────────────┬─────────────────────────────┐                                                                                                         
  │        Filter         │         Returns         │        SQL condition        │
  ├───────────────────────┼─────────────────────────┼─────────────────────────────┤                                                                                                         
  │ VtxoFilterAll         │ All VTXOs               │ No filter                   │
  ├───────────────────────┼─────────────────────────┼─────────────────────────────┤
  │ VtxoFilterSpendable   │ Spendable + recoverable │ NOT spent AND NOT unrolled  │                                                                                                         
  ├───────────────────────┼─────────────────────────┼─────────────────────────────┤                                                                                                         
  │ VtxoFilterSpent       │ Spent + unrolled        │ spent=true OR unrolled=true │                                                                                                         
  ├───────────────────────┼─────────────────────────┼─────────────────────────────┤                                                                                                         
  │ VtxoFilterRecoverable │ Recoverable only        │ Go-side IsRecoverable()     │
  └───────────────────────┴─────────────────────────┴─────────────────────────────┘                                                                                                         
                  
  VtxoFilterSpendable preserves historical behavior — recoverable VTXOs are included
  
                                                                                                                                                                                            
  Test plan       
                                                                                                                                                                                            
  - 22-VTXO pagination test verifying ordering across 5 pages                                                                                                                               
  - Filter tests for all, spendable, spent                                                                                                                                                  
  - Multi-asset VTXO test (subquery correctness)                                                                                                                                            
  - Beyond-last-page returns empty
  - MaxPageSize clamping                                                                                                                                                                    
  - Page{} zero-value returns all                                                                                                                                                           
  - All existing e2e tests updated and passing                                                                                                                                              


@altafan please review                                                             
  

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added pagination and filtering for VTXO listings (page number/size; filters: all, spendable, spent, recoverable).

* **Breaking Changes**
  * VTXO listing API updated to a paginated/filterable call; previous dual-list return removed (clients must adapt).

* **Bug Fixes**
  * Offchain balance now derives from spendable-only VTXOs to avoid extra processing.

* **Tests**
  * Added pagination tests and updated end-to-end tests to use the new VTXO listing API.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/arkade-os/go-sdk/pull/175)

<!-- review_stack_entry_end -->
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


ghost (changes requested):
## Arkana Code Review — `feat/pagination-vtxo-list`

Good refactor overall — unifying the VTXO query surface is the right move. But there are two bugs and one significant downstream breakage concern.

---

### 🔴 BUG: Result ordering within pages is non-deterministic

`store/sql/vtxo_store.go` — `assetVtxoVwRowsToVtxos()` (≈line 432)

The SQL subqueries correctly use `ORDER BY created_at DESC, txid ASC, vout ASC` to select which VTXOs land on which page. But `assetVtxoVwRowsToVtxos` groups rows into a `map[string][]queries.AssetVtxoVw` and then iterates the map:

```go
for _, group := range byOutpoint {
    vtxos = append(vtxos, assetVtxoVwGroupToVtxo(group))
}
```

Go map iteration order is **random**. The `ORDER BY` from SQL is discarded. Callers paginating through results will see VTXOs in random order within each page — this defeats the purpose of pagination.

**Fix**: Use an ordered structure (e.g., a `[]string` of keys preserving insertion order from the row scan) to iterate `byOutpoint` deterministically. Alternatively, sort the final `vtxos` slice by `CreatedAt DESC` before returning.

---

### 🔴 BUG: `VtxoFilterRecoverable` pagination is broken

`store/sql/vtxo_store.go` — `GetVtxos()` (≈line 350-355)

When `filter == VtxoFilterRecoverable` with pagination, the code fetches N rows from `SelectAllVtxosPaginated`, then filters Go-side:

```go
if filter == types.VtxoFilterRecoverable {
    filtered := make([]clientTypes.Vtxo, 0, len(vtxos))
    for _, v := range vtxos {
        if v.IsRecoverable() { filtered = append(filtered, v) }
    }
    return filtered, nil
}
```

This means:
- Request `PageSize=10` → fetch 10 "all" VTXOs → filter → might return 0-10 recoverable VTXOs
- Subsequent pages miss recoverable VTXOs that were counted as "all" but not recoverable on previous pages
- **Pagination invariant is violated**: no correct way to enumerate all recoverable VTXOs page-by-page

**Fix options**:
1. Don't support pagination for recoverable (return error if `PageSize > 0 && filter == VtxoFilterRecoverable`)
2. Move the recoverable check into SQL (e.g., `WHERE swept = true AND spent = false AND expires_at < ?`) — though this may not capture the full `IsRecoverable()` logic
3. Document clearly that recoverable filter always returns all results (ignores pagination)

---

### 🟡 API Design: No total count or "has more" indicator

`types/interfaces.go` — `GetVtxos` returns `([]types.Vtxo, error)`

A pagination API without a total count forces callers to:
- Keep requesting pages until they get an empty result
- Cannot render "page X of Y" or know the total VTXO count without a separate query

Consider returning a result struct: `type VtxoPage struct { Vtxos []Vtxo; Total int; Page int }` or at minimum a `hasMore bool`.

---

### 🟡 Cross-repo breakage (5+ repos affected)

This PR removes `ListSpendableVtxos()` and changes the `ListVtxos()` signature. The following downstream repos will break at compile time:

| Repo | File | Call |
|------|------|------|
| `bancod` | `pkg/contract/taker.go:95` | `ListSpendableVtxos(ctx)` |
| `bancod` | `test/e2e/utils_test.go:144` | `ListVtxos(t.Context())` (old 3-return sig) |
| `demos` | `asset/golang/main.go:100` | `ListSpendableVtxos(ctx)` |
| `asset-demos` | `asset/golang/main.go:100` | `ListSpendableVtxos(ctx)` |
| `layerzero-usdt0-arkade-demo` | `test/e2e_flow_test.go:519` | `ListVtxos(ctx)` (old sig) |
| `introspector-review` | `test/utils_test.go:540`, `contract_id_test.go:151` | `ListVtxos(ctx)` (old sig) |

Please either:
1. Open companion PRs for all affected repos, or
2. Keep backward-compatible wrappers (deprecated) during transition

---

### ✅ What looks good

- Subquery-based pagination for multi-asset VTXOs is correct — paginates at the vtxo level, not the view level. Well-documented rationale in `query_paginated.go`.
- `pageToLimitOffset` correctly handles edge cases (PageNum=0 → 1, clamping to MaxPageSize).
- Removing the manual `Spent || Unrolled` check in `getOffchainBalance` is correct — the SQL filter now handles it.
- Test coverage is solid — `TestVtxoPagination` covers page boundaries, partial pages, multi-asset, MaxPageSize clamping, and spent filter.
- `GetVtxosByOutpoint` rename is clearer than the overloaded `GetVtxos`.

---

**Verdict**: Request changes for the two bugs (ordering + recoverable pagination). The cross-repo breakage needs a migration plan before merge.

⚠️ **Protocol note**: While this PR touches VTXO querying (read path) rather than VTXO spending/signing (write path), incorrect pagination could cause `getOffchainBalance` or `getSpendableVtxos` to return incomplete results if someone accidentally passes a Page with PageSize > 0 in internal callers. Current internal callers all use `Page{}` (fetch all), which is correct. Flagging for human review as the VTXO store interface is protocol-adjacent.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (changes requested):
## Arkana Follow-up Review — `32a3e2f lint`

The new commit is formatting-only (line wrapping for linter). **Both bugs from my previous review are still open:**

### 🔴 Still open: Map iteration destroys SQL ordering

`store/sql/vtxo_store.go:462-466` — `assetVtxoVwRowsToVtxos` still iterates `byOutpoint` map with `for _, group := range byOutpoint`, which randomizes the order that SQL carefully established. Pagination without deterministic ordering is broken.

### 🔴 Still open: `VtxoFilterRecoverable` + pagination returns wrong results

`store/sql/vtxo_store.go:348-370` — Recoverable filter still fetches N "all" rows then filters Go-side, violating pagination invariants.

### 🟡 Still open: Cross-repo breakage plan needed

No companion PRs visible yet for downstream consumers (`bancod`, `demos`, etc.).

---

No new issues introduced by the lint commit. Maintaining request-changes until the two bugs are resolved.

⚠️ Protocol-adjacent — flagging for human review.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

sdk.go
```diff
@@ -84,8 +84,11 @@ type Wallet interface {
 	CompleteUnroll(ctx context.Context, to string) (string, error)
 	OnboardAgainAllExpiredBoardings(ctx context.Context) (string, error)
 	WithdrawFromAllExpiredBoardings(ctx context.Context, to string) (string, error)
-	ListVtxos(ctx context.Context) (spendable, spent []clienttypes.Vtxo, err error)
-	ListSpendableVtxos(ctx context.Context) ([]clienttypes.Vtxo, error)
+	ListVtxos(
+		ctx context.Context,
+		page types.Page,
+		filter types.VtxoFilter,
```
altafan, line 90: wouldn't be nice to keep having the option pattern so the user just calls (?):
```
ListVtxos(ctx) -> gets all vtxos, paginated by default if the response is too big
// Or
ListVtxos(ctx, WithPage(token)) -> gets a specific page given its token
// Or
ListVtxos(ctx, WithSpendableOnly()) -> gets only spendable vtxos, paginated by default if the response is too big
// Or
ListVtxos(ctx, WithSpendableOnly(), WithPage(token)) -> gets a specific page of only spendable vtxos
```

altafan:
took over by #178

PR #178, opened 2026-05-15
Cursor-based pagination

altafan:
This adds cursor-based pagination option to ListVtxos:
- the cursor represents the first vtxo of the very next page.
- page size (Limit) can be customized at will between pages. Defaults to max value (1000) and can assume any value in range [1, 1000].
- to correctly implement pagination, the db always orders vtxos by creation time (+ vtxo oupoint to resolve deterministically if many vtxos have the same creation time) DESC, and it's non-sensitive to insertions happening while retrieving pages.
- adds tests for the new pagination system

Please @sekulicd review

## Summary by CodeRabbit

* **New Features**
  * Filter VTXO listings by spendable/spent status and asset ID.
  * Cursor-based pagination for VTXO listings with opaque next-page tokens.
  * Database index added to speed VTXO queries.

* **Breaking Changes**
  * VTXO listing API now accepts filter/options and returns a paginated cursor instead of separate spendable/spent lists.

* **Tests**
  * Expanded unit and end-to-end tests covering pagination, filters, cursor validation, and stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/go-sdk/pull/178)

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: '1.26.2'
+          go-version: '1.26.3'
       
       - run: go mod tidy
       
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: '1.26.2'
+          go-version: '1.26.3'
 
       - run: go mod tidy
 
--- a/funding.go
+++ b/funding.go
@@ -118,24 +118,6 @@ func (w *wallet) Balance(ctx context.Context) (*types.Balance, error) {
 	}, nil
 }
 
-func (w *wallet) ListSpendableVtxos(ctx context.Context) ([]clienttypes.Vtxo, error) {
-	if err := w.safeCheck(); err != nil {
-		return nil, err
-	}
-
-	return w.store.VtxoStore().GetSpendableVtxos(ctx)
-}
-
-func (w *wallet) ListVtxos(
-	ctx context.Context,
-) ([]clienttypes.Vtxo, []clienttypes.Vtxo, error) {
-	if err := w.safeCheck(); err != nil {
-		return nil, nil, err
-	}
-
-	return w.store.VtxoStore().GetAllVtxos(ctx)
-}
-
 func (w *wallet) NotifyIncomingFunds(
 	ctx context.Context, addr string,
 ) ([]clienttypes.Vtxo, error) {
@@ -157,7 +139,7 @@ func (w *wallet) newOffchainAddress(ctx context.Context) (string, error) {
 func (w *wallet) getOffchainBalance(
 	ctx context.Context,
 ) (*types.OffchainBalance, map[string]uint64, error) {
-	vtxos, _, err := w.store.VtxoStore().GetAllVtxos(ctx)
+	vtxos, err := w.store.VtxoStore().GetSpendableOrRecoverableVtxos(ctx)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -170,10 +152,6 @@ func (w *wallet) getOffchainBalance(
 		amountByExpiration  = make(map[int64]uint64)
 	)
 	for _, vtxo := range vtxos {
-		if vtxo.Spent || vtxo.Unrolled {
-			continue
-		}
-
 		// Classify VTXO by state. Priority: Recoverable > Preconfirmed > default.
 		switch {
 		case vtxo.IsRecoverable():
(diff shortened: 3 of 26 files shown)


ghost (changes requested):
## Arkana Code Review — arkade-os/go-sdk#178

Overall: well-designed pagination. Keyset pagination on `(created_at DESC, txid DESC, vout DESC)` with the limit+1 sentinel pattern is the right approach. The CTE-then-JOIN strategy to avoid cutting multi-asset VTXOs mid-page is smart. Filter hash binding the cursor to its filter set is a nice safety net.

However, there are blocking issues that must be fixed before merge.

---

### 🔴 BLOCKING — Compile Error

**`test/e2e/restore_smoke_test.go:50,94`** — These lines will not compile.

```go
page, err := alice.ListVtxos(ctx, arksdk.WithSpendableOnly())
vtxos := page.Vtxos
```

`ListVtxos` returns `([]clienttypes.Vtxo, string, error)` — three return values, not two. Assigning 3 returns to 2 variables is a Go compile error. And `page` would be a `[]Vtxo` slice, so `page.Vtxos` is nonsensical.

This test was clearly written for a `VtxoPageResult`-based return type that was abandoned in favor of the `(slice, cursor_string, error)` tuple. Fix: use the same pattern as the other e2e tests:

```go
vtxos, _, err := alice.ListVtxos(ctx, arksdk.WithSpendableOnly())
```

---

### 🔴 BLOCKING — Test Assertion Bug

**`funding_opts_test.go:91-94`** — This test will fail:

```go
out, err := decodeCursor("")
require.NoError(t, err)
require.Equal(t, vtxoCursor{}, out)  // ← FAILS
```

`decodeCursor("")` returns `(*vtxoCursor)(nil), nil`. The assertion compares a `vtxoCursor` value with a `*vtxoCursor` nil pointer — `reflect.DeepEqual` returns false because the types differ. Fix either:
- Change assertion to `require.Nil(t, out)`, or
- Change `decodeCursor` to `return &vtxoCursor{}, nil` for empty input (though `nil` is the better semantic for "no cursor").

---

### 🟡 Dead Type

**`types/interfaces.go:29-33`** — `VtxoPageResult` is defined but never referenced anywhere in the PR. It's a leftover from the earlier design that used a struct return. Remove it to avoid confusion.

---

### 🟡 Cross-Repo Breakage — Downstream Consumers

This PR removes `ListSpendableVtxos` and changes the `ListVtxos` signature on the public `Wallet` interface. The following downstream repos will break at compile time on their next `go get`:

| Repo | Files | Old API used |
|------|-------|-------------|
| `bancod` | `pkg/contract/taker.go`, `test/e2e/utils_test.go` | `ListSpendableVtxos`, `ListVtxos` (old sig) |
| `introspector-review` | `test/utils_test.go`, `test/contract_id_test.go` | `ListVtxos` (old sig) |
| `demos` / `asset-demos` | `asset/golang/main.go` | `ListSpendableVtxos` |
| `layerzero-usdt0-arkade-demo` | `test/e2e_flow_test.go` | `ListVtxos` (old sig) |

**Recommendation:** Open companion PRs in affected repos (at minimum `bancod` since it's production) before or alongside this merge, and pin the `go.mod` upgrade together.

---

### ✅ Protocol Correctness Notes

- `getOffchainBalance` switch from `GetAllVtxos` + in-memory filter to `GetSpendableOrRecoverableVtxos` is logically equivalent — the SQL `WHERE spent = false AND unrolled = false` matches the old `if vtxo.Spent || vtxo.Unrolled { continue }` guard. Correct.
- `getSpendableVtxos` in `send.go` and `scheduleNextSettlement` in `init.go` — same rename, same SQL, no behavioral change. Correct.
- `wallet.go` changes from `GetVtxos(ctx, outpoints)` → `GetVtxosByOutpoints(ctx, outpoints)` are pure renames. Correct.
- The `assetVtxoVwRowsToVtxos` fix to preserve insertion order via the `order` slice is necessary for pagination to work — without it, `map` iteration randomizes page order. Good catch.

---

### ✅ SQL Review

- Composite index `idx_vtxo_created_at_txid_vout ON vtxo (created_at DESC, txid DESC, vout DESC)` matches the ORDER BY and cursor predicate exactly. The keyset scan is O(log n + limit). Good.
- Row-value comparison `(created_at, txid, vout) < (?, ?, ?)` is supported in SQLite ≥ 3.15.0 (2016). Safe.
- The `EXISTS` subquery for asset filtering avoids inflating the row count. Correct.
- The CAST wrappers for nullable params are the right sqlc idiom.

---

### ✅ Pagination Design

- Limit+1 sentinel pattern: clean, standard.
- Cursor encodes as `base64url(JSON)` with a filter hash — prevents cursor reuse across different filter sets. Good.
- Filter hash is `sha256[:8]` of status+asset — 64 bits is ample for accidental mismatch detection (this isn't a security boundary).
- `encodeCursor` ignoring the `json.Marshal` error is fine for a struct of primitives.

---

### ⚠️ VTXO Handling — Human Review Required

This PR modifies the VTXO listing and storage interfaces. While the changes are well-structured and the pagination logic is sound, these are **protocol-critical interfaces** per project rules. A human reviewer must sign off before merge.

**Summary:** Fix the two compile/test bugs in restore_smoke_test.go and funding_opts_test.go, remove the dead `VtxoPageResult` type, and coordinate downstream repo updates. The core pagination design is solid.

PR #181, opened 2026-05-18
Pagination PR companion

sekulicd:
Adds e2e coverage for cursor-based ListVtxos pagination, fixes related tests, validates store limits, and clarifies cursor/keyset pagination comments.

### DIFF
--- a/funding_opts_test.go
+++ b/funding_opts_test.go
@@ -64,13 +64,14 @@ func TestVtxoCursorCodec(t *testing.T) {
 		s := encodeCursor(in)
 		out, err := decodeCursor(s)
 		require.NoError(t, err)
-		require.Equal(t, in, out)
+		require.NotNil(t, out)
+		require.Equal(t, in, *out)
 	})
 
-	t.Run("empty cursor decodes to zero value with no error", func(t *testing.T) {
+	t.Run("empty cursor decodes to nil with no error", func(t *testing.T) {
 		out, err := decodeCursor("")
 		require.NoError(t, err)
-		require.Equal(t, vtxoCursor{}, out)
+		require.Nil(t, out)
 	})
 
 	t.Run("malformed base64", func(t *testing.T) {
--- a/sdk.go
+++ b/sdk.go
@@ -84,7 +84,12 @@ type Wallet interface {
 	CompleteUnroll(ctx context.Context, to string) (string, error)
 	OnboardAgainAllExpiredBoardings(ctx context.Context) (string, error)
 	WithdrawFromAllExpiredBoardings(ctx context.Context, to string) (string, error)
+
+	// ListVtxos returns one page of wallet VTXOs plus an opaque cursor for the
+	// next page. The cursor is empty when there are no more results; callers
+	// should pass a non-empty cursor back with WithCursor without parsing it.
 	ListVtxos(ctx context.Context, opts ...ListVtxosOption) ([]clienttypes.Vtxo, string, error)
+
 	Dump(ctx context.Context) (seed string, err error)
 	GetTransactionHistory(ctx context.Context) ([]clienttypes.Transaction, error)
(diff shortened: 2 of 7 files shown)


ghost (approved):
## Review: Pagination PR companion (go-sdk#181)

**Reviewer:** Arkana · **Verdict:** ✅ Approve (with nits)

---

### Summary

Clean companion PR to #178. Adds a `Limit <= 0` guard in the store layer, fixes the sentinel-vs-anchor semantics in comments, changes cursor codec to return `*vtxoCursor` (so empty cursor → `nil` instead of zero-value), updates `restore_smoke_test.go` to the new `ListVtxos` signature, and adds 618 lines of thorough e2e pagination tests.

All changes are on **read-only query paths** (VTXO listing). No protocol-critical code touched — no signing, forfeit, round lifecycle, connector trees, or exit paths.

---

### Findings

#### 1. Comment fix in generated file will be lost on next `sqlc generate` ⚠️

`store/sql/sqlc/queries/query.sql.go:127-129` — The PR updates the header comment to say "trims the sentinel and uses the last returned VTXO as the next cursor anchor", but the **source** file `store/sql/sqlc/query.sql` still has the old wording:

```
-- a has-more sentinel; if SQL returns more than user_limit rows the caller
-- trims the last one and uses its outpoint as the next cursor.
```

Next `sqlc generate` will overwrite the fix. Move the comment correction to `query.sql`.

#### 2. `VtxoPageResult` is dead code (nit)

`types/interfaces.go:36-40` — `VtxoPageResult` struct is defined but never referenced anywhere in the codebase. `GetVtxos` returns `([]types.Vtxo, *Cursor, error)`, not `VtxoPageResult`. If it was from an earlier iteration, remove it; if it's planned for future use, add a `// TODO` explaining why it exists.

#### 3. `walkVtxoPages` has no max-pages safety bound (nit, test-only)

`test/e2e/vtxo_pagination_test.go:410-425` — If a cursor bug causes infinite pagination, this helper loops forever. A `maxPages` guard (e.g., 100) would turn infinite hangs into fast failures. Same applies to the inline pagination loops in `cursor stable when new VTXOs arrive between pages`.

#### 4. `waitAfterVtxoCreatedAt` has no timeout (nit, test-only)

`test/e2e/vtxo_pagination_test.go:399-402` — Loops on `time.Now().Unix() <= vtxo.CreatedAt.Unix()` with no upper bound. If `CreatedAt` were in the future (clock skew, bad test data), this spins forever. A `time.After` guard would be cheap insurance.

---

### What looks good

- **Sentinel/anchor fix is correct.** The old code used the sentinel (extra) row's outpoint as the cursor, which would skip the last real VTXO on each page. The new code correctly uses `vtxos[len(vtxos)-1]` (the last *returned* VTXO) as the anchor after trimming the sentinel. This is the right fix.
- **`decodeCursor` returning `*vtxoCursor`** is the right call — nil semantics are clearer than zero-value-means-absent.
- **`Limit <= 0` guard** in `vtxo_store.go:337` — good defense-in-depth since the option layer already validates `[1, 1000]`.
- **E2e test coverage is excellent.** Tests cover: paginated walk = full result, end-of-pagination nil cursor, limit > dataset, limit changes between pages, status filters, asset filters, multi-asset VTXO integrity across page boundaries, combined filters + pagination, all invalid-input sentinels, cursor-filter mismatch detection, and cursor stability under concurrent writes.
- **Cross-repo impact:** This PR targets the `pagination` branch, not `master`. The `ListVtxos` API signature change is a breaking change for downstream consumers (fulmine, ark-faucet, introspector, bancod, layerzero-demo) but that breakage is introduced by #178, not this PR. This PR only improves correctness and test coverage on top of #178.

---

*No protocol-critical concerns. All findings are nits/minor. Approving because the core logic is correct and well-tested.*

PR #182, opened 2026-05-18
fix for non interactive VHTLC claim

louisinger:
depends on https://github.com/arkade-os/arkd/pull/1068

## Summary by CodeRabbit

* **New Features**
  * Retrieve wallet configuration data
  * Supply additional off-chain send options (including extension packets and Taproot output mapping)
  * Register custom contract handlers via wallet options

* **Improvements**
  * Avoid redundant contract generation for single-key providers

* **Tests**
  * Added tests for custom handler registration and wallet option validation

* **Dependencies**
  * Updated underlying library versions

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/go-sdk/pull/182?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ghost (changes requested):
## 🔍 Aggressive Code Review — arkade-os/go-sdk#182

**Reviewer:** Arkana (automated, protocol-level)
**Verdict:** Request changes + **requires human sign-off** (protocol-critical: contract creation, VTXO spend paths, taproot tree handling)

---

### 🔴 P0 — Dependency on unmerged PR

**go.mod** — `client-lib` bumped to `v0.0.0-20260514151440-4ecc7753e264`, which depends on **arkd#1068** (still OPEN, not merged). If arkd#1068 is rebased, force-pushed, or abandoned, this pseudo-version becomes a dangling commit. **This PR must not merge before arkd#1068 lands.**

The `WithOutputTaprootTree` and `WithExtraPacket` re-exported in `send_opts.go` don't even resolve in the current tree — `go build` would fail against the current `master` of `client-lib`.

---

### 🟡 P1 — Contract idempotent reuse silently discards label

**`contract/manager.go:104` vs `contract/manager.go:119`**

```go
// Line 104 — label is set on the freshly derived contract
contract.Label = o.label

// Lines 119-121 — but then we return the *stored* contract, ignoring the new label
existing, err := m.store.GetContractsByScripts(ctx, []string{contract.Script})
if err == nil && len(existing) > 0 {
    return &existing[0], nil   // ← existing[0].Label may differ from o.label
}
```

If a caller passes `WithLabel("my-vhtlc")` on second invocation, they get back a contract with whatever label was set on first creation. The label option is silently dropped. Either:
- Update the stored label if it differs, or
- Document that labels are first-write-wins for single-key identities.

---

### 🟡 P1 — GetContractsByScripts error silently swallowed

**`contract/manager.go:119`**

```go
existing, err := m.store.GetContractsByScripts(ctx, []string{contract.Script})
if err == nil && len(existing) > 0 {
```

If `GetContractsByScripts` returns an error (DB corruption, closed connection), the code falls through to `AddContract` which will then fail with a UNIQUE constraint error — a confusing double-fault. At minimum, log the error:

```go
existing, err := m.store.GetContractsByScripts(ctx, []string{contract.Script})
if err != nil {
    log.Warnf("%s failed to check for existing contract: %v", logPrefix, err)
} else if len(existing) > 0 {
    return &existing[0], nil
}
```

---

### 🟡 P1 — Single-key detection is heuristic, not explicit

**`contract/manager.go:248-253`**

```go
singleKey := false
if probe, err := m.keyProvider.NextKeyId(ctx, currentKeyId); err == nil {
    if next, err := m.keyProvider.NextKeyId(ctx, probe); err == nil && next == probe {
        singleKey = true
    }
}
```

This infers single-key behavior by probing `NextKeyId` twice and comparing. It works for the current implementation but is fragile — any key provider with side effects, caching, or non-determinism will be misdetected. Consider adding an explicit capability signal:

```go
type keyProvider interface {
    // existing methods...
    IsSingleKey() bool  // or: KeySpaceSize() uint32
}
```

This makes the contract clearer and removes the guesswork. If that's too invasive for this PR, at least add a comment that the detection depends on `NextKeyId` being pure/deterministic.

---

### 🟡 P2 — Breaking `Wallet` interface change

**`sdk.go:47`** — Adding `GetConfigData` to the `Wallet` interface is a compile-breaking change for any downstream code that implements or mocks the interface. The `SendOffChain` variadic addition is backward-compatible (good), but `GetConfigData` is not.

Consumers affected (from cross-repo grep):
- `bancod` — uses `wallet.WalletService` (different interface, likely unaffected)
- `introspector-review` test suite — uses concrete types, likely unaffected
- `demos/asset-demos` — uses concrete types, likely unaffected

Low blast radius, but worth noting in PR description. If there are any mock implementations in downstream test suites implementing `arksdk.Wallet`, they'll break.

---

### 🟢 P3 — No test coverage for single-key paths

Zero tests for:
- `NewContract` idempotent reuse path (line 119)
- `scanContracts` single-key detection + short-circuit (lines 248-312)
- `SendOffChain` with extra options (`WithExtension`, `WithOutputTaprootTree`)

These are protocol-critical code paths affecting VTXO creation and spending. At minimum, add unit tests for:
1. `NewContract` called twice with single-key provider → returns same contract
2. `scanContracts` with single-key provider → doesn't infinite-loop
3. `scanContracts` with single-key provider → correctly finds used contracts

---

### ✅ What looks good

- **`send_opts.go`**: Clean re-export pattern. Type alias (`= client.SendOption`) avoids wrapper indirection. Good.
- **`wallet.go:358-363`**: `GetConfigData` delegates correctly with nil guard. Consistent with other accessor methods that skip `safeCheck`.
- **`send.go:60`**: `extraOpts` appended after built-in opts — correct ordering (caller opts override defaults).
- **`contract/manager.go:309-311`**: `break scan` after single-key batch is correct — prevents infinite loop.
- **Variadic `SendOffChain` signature**: Backward-compatible. All existing callers compile without changes.

---

### 🚨 Protocol-Critical Flag

This PR modifies:
- **Contract creation** — idempotent reuse for single-key identities
- **Contract scanning** — gap-limit bypass for single-key providers
- **VTXO spending** — taproot tree attachment on SendOffChain outputs

All three are **protocol-critical paths**. Per policy: **human review required before merge**, even if the above issues are addressed.

---

**TL;DR**: Fix the swallowed error (P1), decide on label behavior (P1), add tests for single-key paths, and wait for arkd#1068 to merge. Then get a human to sign off.

🤖 _Reviewed by Arkana_

ghost (commented):
## 🔍 Incremental Review — New commit `14a38c27` (extra contract handlers)

**Reviewer:** Arkana (automated, protocol-level)
**Scope:** Only reviewing changes since my last review (2026-05-18). New commit: `14a38c27` ("allow to pass extra contract handlers") + `63099f4a` (dep bump).

---

### Status of previous findings

| # | Finding | Status |
|---|---------|--------|
| 🔴 P0 | arkd#1068 dependency still unmerged | **Still open.** go.mod now points to `ae820b3a4d71` (2026-05-21) but arkd#1068 remains OPEN. Same risk applies — this PR must not merge before arkd#1068 lands. |
| 🟡 P1 | Swallowed `GetContractsByScripts` error | **Unaddressed.** Code unchanged. |
| 🟡 P1 | Label silently discarded on idempotent reuse | **Unaddressed.** Code unchanged. |
| 🟡 P1 | Single-key detection heuristic | **Unaddressed.** Code unchanged. |
| 🟡 P2 | Breaking `Wallet` interface (`GetConfigData`) | **Unaddressed.** Still breaks `bancod`'s `mockArkClient` (embeds `arksdk.Wallet`). |
| 🟢 P3 | No tests for single-key paths | **Unaddressed.** |

---

### Review of new code

#### ✅ `contract/manager.go` — ExtraHandlers merge (lines 41-52)

Clean. Nil guard is good. Built-in handlers constructed first, extras merged last — correct override semantics. Map iteration order irrelevant since we're inserting into another map.

#### ✅ `contract/types.go` — ExtraHandlers field + docs

Well-documented. Clearly states override behavior for default/boarding types.

#### 🟡 P2 — Map aliasing in `WithContractHandlers`

**`wallet_opts.go:110-122`**

```go
func WithContractHandlers(extra map[types.ContractType]handlers.Handler) WalletOption {
    return func(o *walletOptions) error {
        // ...validation...
        o.extraContractHandlers = extra  // ← stores caller's map reference
        return nil
    }
}
```

The map is stored by reference, not copied. If the caller mutates the map after calling `WithContractHandlers` (e.g. adds entries, sets values to nil), the wallet's handler registry silently changes — bypassing the nil-handler validation. Defensive copy:

```go
o.extraContractHandlers = maps.Clone(extra)
```

`maps` is already imported in `contract/manager.go`, so the dependency exists. Minor but worth fixing since this is a public API that validates invariants on the map contents.

#### ✅ `wallet.go` — Plumbing

`extraContractHandlers` correctly wired in both `NewWallet` and `LoadWallet`. Passed through to `contract.NewManager` in `init.go:108`. No issues.

#### ✅ `contract/manager_extra_handlers_test.go` — Good coverage

Tests cover: custom type registration, NewContract delegation, GetHandler lookup, nil handler rejection. Clean test helpers. Uses `t.Context()` (Go 1.24+) — fine for this codebase.

#### ✅ `wallet_opts_test.go` — Good coverage

Tests cover: valid registration, empty map, nil entry, double-set. All expected error paths tested.

---

### 🚨 Protocol-Critical Flag (reiterated)

Previous findings on contract creation idempotency, single-key scanning, and taproot tree handling remain unaddressed. These are protocol-critical paths — **human review still required before merge**.

---

**TL;DR for this increment:** The ExtraHandlers feature is well-implemented with good test coverage. One minor issue: defensive-copy the handler map in `WithContractHandlers` to prevent post-validation mutation. Previous P0/P1 findings remain open.

🤖 _Reviewed by Arkana_

PR #186, opened 2026-05-22
Contract registry: allow registring custom contract handlers

### LINKED ISSUE #151: Registry: support contracts defined by others
Our sdk supports natively default, delegate and vhtlc contracts. It would be super powerful to allow users to define their own contracts and to have a way to add them to those supported by the client's contract manager.

This is exactly the purpose of the contract registry that would come with 3 contract types supported by default and the possibility to add custom ones defined by users.

Ref: https://github.com/arkade-os/ts-sdk/blob/master/src/contracts/handlers/registry.ts

altafan:
Closes #151.

This adds a new `contract.Registry` interface and related implementation to allow users to provide their own contract handlers at wallet creation with the newly introduced `WithContractHandler` option.

Please @bitcoin-coder-bob @sekulicd @louisinger review

## Summary by CodeRabbit

* **New Features**
  * Register and use custom contract handlers when creating a wallet; manager exposes a sealed runtime handler registry and accepts handler options at construction.

* **Bug Fixes / Reliability**
  * Stronger validation and clearer errors for nil/typed-nil, empty, or duplicate handler registrations; built-in handlers use a shared cached client.

* **Documentation**
  * Clarified handler caching, concurrency model, and updated guidance for adding custom contract types.

* **Tests**
  * Added unit and end-to-end tests for custom handlers, registry behavior, concurrency, and handler validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/go-sdk/pull/186?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ghost (changes requested):
## Code Review — Contract Registry (#186)

**Reviewer: Arkana**
**Verdict: Request changes** (1 protocol concern, 2 bugs, 2 design issues)

Overall this is a well-structured PR. The immutability-over-mutex approach is sound and the validation is thorough. Test coverage is good. No cross-repo breakage detected — `bancod` (the only downstream consumer) uses `GetContracts`/`GetHandler`/`WithScripts`, all of which are preserved.

---

### 🔴 Protocol Concern — Custom handlers silently participate in ScanContracts

**`contract/manager.go:64-76`** (new code via diff)

Custom handlers registered via `WithHandler` are scanned in `ScanContracts` using the default `findUsedContracts` (indexer/offchain) probe for any non-boarding type. If a custom contract type requires a different "has this contract been used?" strategy (e.g., an on-chain lookup like boarding, or a completely custom path), there is **no way to configure this**.

A buggy or slow custom handler will also block the entire wallet unlock flow since `ScanContracts` runs in the background goroutine in `Unlock()` (init.go) and errors propagate to `syncCh`.

**Recommendation:** At minimum, document this clearly in the `WithHandler` godoc. Ideally, consider either:
- Adding a `FindUsedFunc` to the `Handler` interface (or a separate optional interface), or
- Allowing custom handlers to opt out of `ScanContracts` via a `ScanSkip` sentinel.

This is protocol-adjacent — a custom handler that incorrectly reports contracts as "unused" could lead to missed VTXOs.

---

### 🟡 Bug — Duplicated validation logic will drift

The typed-nil / nil handler validation is copy-pasted in **three places**:

1. `contract/opts.go:358-368` (`WithHandler`)
2. `contract/registry.go:80-96` (`validateHandler`)
3. `wallet_opts.go:148-160` (`WithContractHandler`)

Location 2 uses the shared `validateHandler` helper. Locations 1 and 3 inline the same reflect logic. If the validation rules ever change, the inline copies will be forgotten.

**Fix:** `WithHandler` (opts.go) should call `validateHandler` — it already imports the same package. `WithContractHandler` (wallet_opts.go) can't easily call the contract-package helper due to import direction, but the reflect block should at minimum be extracted to a local helper to reduce duplication within the file. Alternatively, move `validateHandler` to a shared internal package.

---

### 🟡 Bug — `NewRegistry` exported with unexported return type

**`contract/registry.go:38`**

```go
func NewRegistry(builtIns, customs map[types.ContractType]handlers.Handler) (*registry, error) {
```

`NewRegistry` is exported but returns `*registry` (unexported). External callers can call it but can't name the return type, making it awkward to use outside tests. If this is intentional (test-only export), document it. If it's meant to be a public API, export `Registry` as the return type:

```go
func NewRegistry(...) (Registry, error) {
```

---

### 🟢 Minor — Test name typos

**`wallet_opts_test.go:54,60`**: `"WitContralHandler"` → should be `"WithContractHandler"`.

---

### 🟢 Minor — Map iteration in init.go

**`init.go:103-105`**:
```go
for t, h := range w.customHandlers {
    mgrOpts = append(mgrOpts, contract.WithHandler(t, h))
}
```

Map iteration order is non-deterministic. Since `WithHandler` rejects duplicates, this is *correct* but means the order of `ManagerOption` application varies across runs. Not a bug today, but worth a comment if option ordering ever matters.

---

### ✅ What looks good

- **Immutability model**: Removing the `sync.RWMutex` in favor of sealing the registry at construction is cleaner and eliminates a class of race conditions. The doc update in `doc.go` accurately reflects this.
- **Validation depth**: Catching typed-nil via reflect is a nice touch — common Go footgun.
- **Test coverage**: Unit tests for registry, opts, manager integration, and an e2e test. Edge cases (empty type, nil, typed-nil, duplicate, builtin collision) all covered.
- **No cross-repo breakage**: `bancod` doesn't use `GetSupportedContractTypes` or call `NewManager` directly. Safe.
- **Interface evolution**: Replacing `GetSupportedContractTypes` with `Registry()` is a better abstraction — gives callers both type enumeration and handler lookup through one accessor.

---

**⚠️ Flagging for human review** — this PR modifies the contract manager interface and how custom handlers participate in contract scanning. While the code is well-written, the protocol implications of custom handlers in `ScanContracts` (missed VTXOs, wallet unlock blocking) warrant human sign-off.

ghost (commented):
## Incremental Review — `7428b59a` ("Fix race condition")

**Reviewer: Arkana** (follow-up to previous review)

Only one file changed: `test/e2e/custom_handler_test.go` — adds `<-arkClient.IsSynced(t.Context())` to drain background goroutines before `t.Cleanup` invokes `Stop`.

### ✅ The fix is correct

The race was: `Stop()` runs via `t.Cleanup` while `ScanContracts` bg goroutines are still active. Draining `IsSynced` ensures sync completes (success or error) before teardown. The comment accurately explains that a sync error is expected since the custom handler produces non-standard scripts.

### ⚠️ Minor concern — `IsSynced` doesn't respect context cancellation

`IsSynced(ctx)` ignores the `ctx` parameter — the returned channel is only signaled by the sync completing or `Stop()` broadcasting. If `ScanContracts` were to hang indefinitely (e.g., a misbehaving custom handler), this drain would block the test forever. In practice it's fine here because the custom handler is trivial and the indexer rejection is fast, but worth noting for future test authors.

No new issues. **Previous review findings still stand** (duplicated validation logic, exported `NewRegistry` returning unexported type, test name typos). This commit doesn't address those — which is fine, it's a targeted race fix.

ghost (commented):
## Incremental Review — `9a99360` ("Fix after review")

**Reviewer: Arkana** (follow-up #2)

Only `test/e2e/custom_handler_test.go` changed. This commit adds a 30-second `context.WithTimeout` around the `IsSynced` drain, with a `select` that fails the test on timeout instead of hanging CI.

### ✅ This directly addresses my previous concern

The pattern is correct:

```go
syncCtx, cancel := context.WithTimeout(t.Context(), 30*time.Second)
defer cancel()
select {
case <-arkClient.IsSynced(syncCtx):
case <-syncCtx.Done():
    require.FailNowf(t, "sync drain timeout", "IsSynced did not return: %v", syncCtx.Err())
}
```

- `context.WithTimeout` properly bounds the wait.
- `defer cancel()` prevents context leak.
- `require.FailNowf` correctly halts the test (vs `t.Error` which would continue).
- 30s is generous enough for CI but tight enough to catch real hangs.

### Remaining items from original review (not addressed in this commit)

These are non-blocking but still worth tracking:

1. **Duplicated validation** — `WithHandler` (opts.go) and `NewRegistry` (registry.go) both validate empty type + nil handler. The registry comment acknowledges this is intentional defense-in-depth. Acceptable.
2. **`NewRegistry` exported, returns `*registry` (unexported)** — minor API hygiene; callers outside the package can call it but can't name the return type. Low priority.
3. **Test name typos** — `TestWithHandler_DuplicateType`/similar test names from earlier review are fine as-is.

No new issues found. Previous `CHANGES_REQUESTED` findings are either resolved (this commit) or acknowledged as non-blocking.

ghost (approved):
**Reviewer: Arkana** (follow-up #3 — commit `feeb8178`)

## ✅ LGTM

Changes since `9a99360` are exactly what was requested:

1. **`contract/manager.go:24`** — `registry` field now stores `Registry` (interface) instead of `*registry` (concrete). ✔️
2. **`contract/registry.go:35`** — `NewRegistry` return type changed from `*registry` → `Registry`. ✔️  
3. **`wallet_opts_test.go:55,62`** — Typo fix in test names (`WitContralHandler` → `WitContractHandler`). ✔️

No new logic, no new risk surface. All prior feedback addressed. Approving.

> ⚠️ **Protocol-critical reminder**: This PR touches contract handler registration which feeds into VTXO lifecycle. Human approval still required before merge.

bitcoin-coder-bob:
something in the README in the `Advanced Usage` section for `WithContractHandler` seems valuable to add.


```
 Custom Contract Handlers
  
  The contract manager ships with built-in handlers for the default (offchain) and boarding contract types. You can teach it additional contract types (for example a VHTLC or a delegation script) by
  registering your own handlers.Handler at wallet construction:

  import "github.com/arkade-os/go-sdk/contract/handlers"

  // myHandler implements contract/handlers.Handler. The key method is
  // NewContract, which derives the script, address, and params for a given key.
  // See contract/handlers/default for a reference implementation.
  var myHandler handlers.Handler = NewVHTLCHandler(...)

  wallet, err := arksdk.NewWallet(
      datadir,
      arksdk.WithContractHandler("vhtlc", myHandler),
  )

  Once registered, create and look up contracts of the custom type through the usual manager API:

  mgr := wallet.ContractManager()

  // Discover the registered types (built-ins plus your custom ones).
  supported := mgr.Registry().SupportedTypes() // [boarding default vhtlc]

  // Create a contract using your handler.
  c, err := mgr.NewContract(ctx, "vhtlc")
```

bitcoin-coder-bob:
Regarding concurrency here. This PR removes the mutex from contractManager. On master NewContract and ScanContracts both did m.mu.Lock() for the whole function which serialized the get latestcontract -> derive next key -> AddContract sequence. Now there's no locking left in contract/manager.go at all. And nothing else picks up the slack. In store/sql/contract_store.go AddContract (line 31) and GetLatestContract (line 121) don't take any lock, the store's lock only gets used in Clean (line 152). The reason I think this matters is it's reachable from normal usage not just if someone calls things concurrently on purpose. Unlock kicks off ScanContracts in a background goroutine (init.go lines 138-147) and returns before it's done so if the caller calls NewOffchainAddress/NewBoardingAddress right after unlock you've got the background scan and the foreground NewContract both running that same derive and store path on the same store with nothing serializing them. It is a time of check/time of use race not a memory race so the race detector won't catch it and there is no corruption. The script primary key keeps the db consistent so the worst case is one of the callers gets a spurious "contract already exists" error and some wasted work. But that's a regression and the old code specifically prevented this. 

Simplest fix is just put the write lock back around NewContract and ScanContracts. Or wrap the read/derive/insert in a single db transaction. If the manager isn't meant to be concurrency safe anymore then lets just say so somewhere and serialize at the call sites instead.

bitcoin-coder-bob:
Here is some AI feedback worth considering:

1. Duplicated handler validation. Urgency: low. The nil and typed-nil reflect check exists in two places. `contract/registry.go` `validateHandler` and `wallet_opts.go` `WithContractHandler` are near identical copies and will drift. `validateHandler` is unexported so either export a shared validator or route the wallet option through `contract.WithHandler`.

2. `NewRegistry` is exported but can't actually be used as an extension point. Urgency: low. The manager always builds its own registry from `WithHandler` options so an externally constructed `Registry` can't be injected. It looks like unintended public surface. Unexporting it to `newRegistry` would also match the `WithHandler` doc comment which already says "detected later, inside newRegistry".

3. `ScanContracts` scans custom types through the offchain indexer path. Urgency: medium. It iterates `registry.SupportedTypes()` and uses `findUsedContracts` (indexer `GetVtxos`) for everything that isn't boarding including custom types. Custom scripts may not be indexable that way. The e2e test in `test/e2e/custom_handler_test.go` even notes arkd rejects them and a sync error is expected. So registering a custom handler can make background sync error out and there is no way to opt a type out of scanning. Worth a design note or an opt-out.

4. No test that `NewContract` with a custom type dispatches to the custom handler and persists. Urgency: medium. Current coverage is registration plus `GetHandler` lookup plus `SupportedTypes` ordering. The e2e test deliberately avoids creating a custom contract. The `mockHandler` already returns a deterministic script and address so a unit test that drives `NewContract` through a custom type and asserts it is stored is cheap and would also exercise `newContract` and scan with a non built-in type.

5. No concurrency test for `NewContract`. Urgency: low. This ties to the mutex removal issue. If concurrent allocation is meant to be safe add a `-race` test that calls `NewContract` from several goroutines and asserts unique scripts and no spurious errors. If it is not meant to be safe document the constraint.

6. Security. Urgency: low. No real vulnerabilities. The custom handler API is a trust equivalent extension point. A buggy handler can produce a script the identity cannot sign for which sends funds to an unspendable address. That is a footgun not a vuln but a doc warning on `WithContractHandler` and `WithHandler` that handlers must produce signable scripts is worth adding. The reflect typed-nil guards are good defensive practice.

bitcoin-coder-bob (changes requested):
overall the changes in this PR LGTM and make sense, left a few small items requesting changes

ghost (commented):
**Reviewer: Arkana** (follow-up #4 — commits `930e2814`, `6de0ed9b`)

Focused on delta since `feeb8178`. Two items:

---

### 1. 📄 Stale doc.go — concurrency section contradicts code (low)

`contract/doc.go:82-86` still says:

> The handler map is sealed at NewManager time … **There is therefore no mutex on the manager** — registry reads are concurrent-safe by virtue of immutability.

But `930e2814` restored `sync.RWMutex` on `contractManager` and acquires it in `ScanContracts`, `NewContract`, `GetContracts`, `Clean`, and `Close`. The doc should be updated to reflect the current locking strategy — something like:

> The handler map (registry) is sealed at NewManager time and needs no lock. The manager guards store and key-provider access with an `sync.RWMutex`: mutations (`NewContract`, `ScanContracts`, `Clean`, `Close`) hold the write lock; reads (`GetContracts`) hold the read lock.

---

### 2. ⚠️ `ScanContracts` holds write lock during network I/O (medium)

`contract/manager.go:69-84`: `ScanContracts` acquires `m.mu.Lock()` at the top and holds it for the entire scan loop, including:
- `findUsedContracts` → `m.indexer.GetVtxos()` (network call)
- `findUsedBoardingContracts` → `m.explorer.GetTxs()` per address + `time.Sleep(time.Second)` every 20 calls

With a gap limit of 20 and many boarding addresses, the write lock can be held for **minutes**, blocking all `GetContracts` reads.

The lock is correctly needed around the key-derivation + store-write sequence in `newContract` to prevent duplicate key indices. But in `ScanContracts`, consider:
- Moving the lock **inside** `scanContracts`, acquiring it only around `m.store.AddContract` (the persist loop at the bottom)
- Or narrowing it to lock-per-batch rather than lock-for-entire-scan

Not blocking — the scan is infrequent (wallet unlock / explicit recovery) — but worth a TODO for when concurrent wallet access matters.

---

### Other observations (all clean, no action needed):

- **`validateHandler` → `utils.ValidateHandler`**: Good dedup. Both `contract/opts.go:18` and `wallet_opts.go:122` now share the same validation path. Logic is identical to before.
- **`NewRegistry` → `newRegistry` + public wrapper**: `NewRegistry` now delegates to `newRegistry` which returns `*registry`. Clean refactor, preserves backward compat for any external callers.
- **`mu: sync.RWMutex{}`** in struct literal (`manager.go:63`): Zero-value init is unnecessary but harmless.
- **Concurrency test** (`manager_test.go:87-117`): Good coverage — 10 goroutines racing on `NewContract`, verifying unique scripts and key IDs.
- **Custom handler dispatch test** (`manager_test.go:561-585`): Properly verifies end-to-end: custom type → handler dispatch → store persistence → query by type.

LGTM pending the doc fix. The lock granularity is a note for future improvement, not a blocker.

bitcoin-coder-bob:
from my prior comment:

```
ScanContracts scans custom types through the offchain indexer path. Urgency: medium. It iterates registry.SupportedTypes() and uses findUsedContracts (indexer GetVtxos) for everything that isn't boarding including custom types. Custom scripts may not be indexable that way. The e2e test in test/e2e/custom_handler_test.go even notes arkd rejects them and a sync error is expected. So registering a custom handler can make background sync error out and there is no way to opt a type out of scanning. Worth a design note or an opt-out.
```

this was not addressed in recent commits but i think should be.

custom handler should not be able to silently kill the wallet's tx listeners and periodic refresh

 I suggest one of:

  1. Make per-type scan non-fatal. In `ScanContract` log and continue on a typeds error instead of return err so one type can't kill the whole loop.
  2. Or do a scan opt out. Let `WithHandler` mark a type as skip scan so that 0custom types dont hit the indexer at all by default

ghost (commented):
**Reviewer: Arkana** (follow-up #5 — commits `4f93e84`, `b5137a6`, `7011241`)

Focused on the three new commits: README example, doc.go update, and `handlerSanityCheck` in `ValidateHandler`.

---

### 1. 🔴 `handlerSanityCheck` has no context timeout — can hang forever

**`internal/utils/utils.go:73`** — `handlerSanityCheck` calls `h.NewContract(context.Background(), fakeId)`. A well-behaved custom handler may do network I/O in `NewContract` (e.g., fetching ASP pubkey or network params). If the server is unreachable, this blocks wallet construction indefinitely with no way to cancel.

```go
// current
c, err := h.NewContract(context.Background(), fakeId)

// suggested
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()
c, err := h.NewContract(ctx, fakeId)
```

Severity: **medium**. This runs during `Unlock` → `NewManager` → `WithHandler` → `ValidateHandler`. A hung handler = a hung wallet.

---

### 2. 🟡 Sanity check doesn't validate `Script` or `Address` are non-empty

**`internal/utils/utils.go:84-88`** — The check verifies `c.Type` matches but doesn't verify that the returned contract has a non-empty `Script` (or `Address`). A handler that returns `Type: "custom"` with an empty script passes validation but produces an unusable contract that will fail at persistence or signing time.

Add after the type check:
```go
if c.Script == "" {
    return fmt.Errorf("custom handler NewContract returned empty Script for type %q", t)
}
```

---

### 3. 🟡 README example has wrong `SupportedTypes` comment

**`README.md:410`** — The example registers type `"custom"` but the comment says:
```go
supported := mgr.Registry().SupportedTypes() // [boarding default vhtlc]
```
Should be `// [boarding custom default]` (alphabetically sorted, matching the registered type). Minor, but incorrect docs for a new API will confuse early adopters.

---

### 4. 🟡 doc.go is inconsistent between disk and diff

**`contract/doc.go:77-82`** — The diff updates the Concurrency section to document the new `sync.RWMutex` on the manager, but the Concurrency section references appear in both old and new form in the PR. Confirm the final state reads correctly — the current doc.go text should mention the RWMutex, not the old "no mutex" language.

---

### 5. 🔴 @bitcoin-coder-bob's `ScanContracts` concern remains unaddressed

This was flagged in the previous review cycle and again by bob in comments. Custom types flow through `ScanContracts` via the offchain indexer path (`findUsedContracts`). Custom scripts that aren't indexable by arkd will error, and that error propagates up and **kills the wallet's background sync loop**. This is the single biggest risk in the PR — a user registers a custom handler and their wallet's periodic refresh breaks.

Minimum viable fix: make per-type scan errors non-fatal (log + continue), or let `WithHandler` mark a type as `skipScan`. This needs resolution before merge.

---

### Test coverage: ✅ Good

The new `mockedHandler` + testify mock approach is solid. Each handler method failure path is individually tested. The "wrong type" case is covered. Well done.

---

**Summary**: The sanity check is a good addition but needs a context timeout and a Script emptiness check. The ScanContracts issue remains the blocking concern. README comment is a quick fix.

ghost (commented):
**Reviewer: Arkana** (follow-up #5 — commits `4f93e84`, `b5137a6`, `7011241`)

Focused on the three new commits: README example, doc.go update, and `handlerSanityCheck` in `ValidateHandler`.

---

### 1. 🔴 `handlerSanityCheck` has no context timeout — can hang forever

**`internal/utils/utils.go:73`** — `handlerSanityCheck` calls `h.NewContract(context.Background(), fakeId)`. A well-behaved custom handler may do network I/O in `NewContract` (e.g., fetching ASP pubkey or network params). If the server is unreachable, this blocks wallet construction indefinitely with no way to cancel.

```go
// current
c, err := h.NewContract(context.Background(), fakeId)

// suggested
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()
c, err := h.NewContract(ctx, fakeId)
```

Severity: **medium**. This runs during `Unlock` → `NewManager` → `WithHandler` → `ValidateHandler`. A hung handler = a hung wallet.

---

### 2. 🟡 Sanity check doesn't validate `Script` or `Address` are non-empty

**`internal/utils/utils.go:84-88`** — The check verifies `c.Type` matches but doesn't verify that the returned contract has a non-empty `Script` (or `Address`). A handler that returns `Type: "custom"` with an empty script passes validation but produces an unusable contract that will fail at persistence or signing time.

Add after the type check:
```go
if c.Script == "" {
    return fmt.Errorf("custom handler NewContract returned empty Script for type %q", t)
}
```

---

### 3. 🟡 README example has wrong `SupportedTypes` comment

**`README.md:410`** — The example registers type `"custom"` but the comment says:
```go
supported := mgr.Registry().SupportedTypes() // [boarding default vhtlc]
```
Should be `// [boarding custom default]` (alphabetically sorted, matching the registered type). Minor, but incorrect docs for a new API will confuse early adopters.

---

### 4. 🔴 @bitcoin-coder-bob's `ScanContracts` concern remains unaddressed

This was flagged in the previous review cycle and again by bob in comments. Custom types flow through `ScanContracts` via the offchain indexer path (`findUsedContracts`). Custom scripts that aren't indexable by arkd will error, and that error propagates up and **kills the wallet's background sync loop**. This is the single biggest risk in the PR — a user registers a custom handler and their wallet's periodic refresh breaks.

Minimum viable fix: make per-type scan errors non-fatal (log + continue), or let `WithHandler` mark a type as `skipScan`. This needs resolution before merge.

---

### Test coverage: ✅ Good

The new `mockedHandler` + testify mock approach is solid. Each handler method failure path is individually tested. The "wrong type" case is covered. Well done.

---

**Summary**: The sanity check is a good addition but needs a context timeout and a Script emptiness check. The ScanContracts issue remains the blocking concern. README comment is a quick fix.

ghost (commented):
**Reviewer: Arkana** (follow-up #6 — commit `dc422f1`)

Reviewed the delta since `7011241`. Three changes, all addressing prior feedback:

### ✅ Issues fixed in this commit

1. **`internal/utils/utils.go:98-100`** — Nil contract check in `handlerSanityCheck`. Prevents nil deref on `c.Type` at line 101. Good.
2. **`README.md:388-416`** — Fixed the type mismatch (`"vhtlc"` → `"custom"`), added the unspendable-contract warning, corrected `SupportedTypes` comment. All consistent now.
3. **`wallet_opts_test.go:215-216`** — `GetKeyRef` mock returns `(nil, nil)` instead of `errors.New("not implemented")`, so the sanity check passes. Correct fix.

### 🟡 Still outstanding from previous reviews

Two items from follow-up #5 remain unaddressed:

1. **`internal/utils/utils.go:94`** — `handlerSanityCheck` still uses `context.Background()` with no timeout. A custom handler doing network I/O in `NewContract` blocks wallet construction indefinitely. Please add a `context.WithTimeout` (10-30s is fine).

2. **`internal/utils/utils.go:98-104`** — The nil check is good, but there's still no validation that `c.Script != ""`. A handler returning `Type: "custom"` with an empty script passes the sanity check but produces an unusable contract. One-liner to add after line 100:
   ```go
   if c.Script == "" {
       return fmt.Errorf("custom handler NewContract returned empty Script for type %q", t)
   }
   ```

These are not blockers for the core design, but they harden the validation path you've already built. No new architectural concerns — the PR is converging well.

---
🤖 Arkana · automated review · model: opus

altafan:
> from my prior comment:
> 
> ```
> ScanContracts scans custom types through the offchain indexer path. Urgency: medium. It iterates registry.SupportedTypes() and uses findUsedContracts (indexer GetVtxos) for everything that isn't boarding including custom types. Custom scripts may not be indexable that way. The e2e test in test/e2e/custom_handler_test.go even notes arkd rejects them and a sync error is expected. So registering a custom handler can make background sync error out and there is no way to opt a type out of scanning. Worth a design note or an opt-out.
> ```
> 
> this was not addressed in recent commits but i think should be.
> 
> custom handler should not be able to silently kill the wallet's tx listeners and periodic refresh
> 
> I suggest one of:
> 
> 1. Make per-type scan non-fatal. In `ScanContract` log and continue on a typeds error instead of return err so one type can't kill the whole loop.
> 2. Or do a scan opt out. Let `WithHandler` mark a type as skip scan so that 0custom types dont hit the indexer at all by default

we def want to include "custom contracts" in the rescan process, and we def want that operation to fail if something goes wrong for a custom handler. But also, we def don't want to discover a custom handler is buggish in manager's ScanContracts or NewContract, and therefore we added extra checks in `utils.ValidateHandler` so we create a dummy contract and test all handlers' apis. The util function is used both in the wallet option `WithCustomHandler` and in the contract manager, so we'll spot bugs in custom handlers early, at wallet setup (`NewWallet`/`LoadWallet`)

ghost (commented):
**Reviewer: Arkana** (follow-up #7 — commit `d060027`)

Reviewed delta since `dc422f1`. One production change, rest is test refactoring.

### ✅ Production code

**`internal/utils/utils.go:105-106`** — Empty script check added. This was outstanding item #2 from follow-up #6. Good.

### ✅ Test refactoring

The switch from hand-rolled `mockHandler` to `testify/mock`-based `mockedHandler` is clean:
- `mockHandler()` helper wires default successful responses; tests override via specific `.On()` calls before calling the helper (first-match-wins). Correct pattern.
- `h.AssertCalled()` in `TestManagerWithCustomHandlers` verifies dispatch. Good.
- `vhtlc`/`delegate` → `custom`/`custom-1`/`custom-2` naming is consistent across all test files.
- Expanded `TestNewRegistry` covers builtins-only, customs-only, and merged scenarios.

No issues found in this commit.

### 🟡 One item still outstanding

**`internal/utils/utils.go:95`** — `handlerSanityCheck` still calls `h.NewContract(context.Background(), ...)` with no timeout. A misbehaving custom handler doing network I/O blocks `NewWallet`/`LoadWallet` indefinitely. One-liner fix:

```go
ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
defer cancel()
c, err := h.NewContract(ctx, fakeId)
```

This is a minor hardening point, not a blocker. The PR is in good shape.

---
🤖 Arkana · automated review · model: opus

PR #189, opened 2026-05-24
fix: poll arkd until ready instead of failing on first connection error

### LINKED ISSUE #188: `make regtest` sometimes fails because it talks to arkd before it's ready
when you run `make regtest`, it starts the arkd server and then immediately tries to talk to it. The trouble is that the server takes a few seconds to fully wake up after starting, and the old setup code only waited a fixed 3 seconds before asking it a single question. 

If the server wasn't ready in those 3 seconds, which happens on slower machines or on a fresh build, it would refuse the connection (the "connection reset by peer" error), and the whole command would give up instead of just waiting a little longer.

Image

Dunsin-cyber:
fixes #188

**before** 
go-sdk--startupp-err

**after**
go-sdk-startup-err-fixed

## Summary by CodeRabbit

* **Chores**
  * Enhanced test setup with improved service readiness verification and error reporting for better reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/go-sdk/pull/189?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/test/docker/setup.go
+++ b/test/docker/setup.go
@@ -34,8 +34,9 @@ func setupArkd() error {
 	}
 
 	fmt.Println("waiting for arkd to be ready...")
-	url := fmt.Sprintf("%s/v1/admin/wallet/status", adminUrl)
-	status, err := getStatusWithRetry(adminHttpClient, url, 30, 2*time.Second)
+	status, err := waitForStatus(adminHttpClient, 30, 2*time.Second, func(*statusResp) bool {
+		return true
+	})
 	if err != nil {
 		return err
 	}
@@ -59,7 +60,7 @@ func setupArkd() error {
 	}
 
 	fmt.Println("getting wallet seed...")
-	url = fmt.Sprintf("%s/v1/admin/wallet/seed", adminUrl)
+	url := fmt.Sprintf("%s/v1/admin/wallet/seed", adminUrl)
 	seed, err := get[seedResp](adminHttpClient, url, "seed")
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


ghost (approved):
## ✅ Approve — good improvement to test infra reliability

**Scope:** `test/docker/setup.go` only — test harness, no production SDK code, no protocol-critical paths, no public API changes. Safe to merge.

**What it does right:**
- Replaces the fragile `time.Sleep(3s)` + single-shot call with a proper retry loop (30 × 2s = 60s max)
- Makes `waitUntilReady` tolerant of transient connection errors instead of failing immediately
- Bounded retries instead of the old infinite `ticker` loop — prevents CI from hanging forever
- Clear logging on each retry attempt

**Minor suggestions (non-blocking):**

1. **Duplication between `getStatusWithRetry` and `waitUntilReady`** — these are nearly identical (same endpoint, same interval, same attempt count). `waitUntilReady` could call `getStatusWithRetry` internally and then check the `Initialized && Unlocked && Synced` fields, or both could share a common retry helper. Not blocking since this is test infra, but worth a quick cleanup.

2. **Unnecessary sleep after final attempt** — both functions `time.Sleep(interval)` after the last failed attempt before returning the error. Minor, but a `if i < attempts { time.Sleep(...) }` guard would shave 2s off failure cases.

3. **`waitUntilReady` drops the last error** (`setup.go:153`) — if all 30 attempts return HTTP errors, the error message is `"arkd wallet did not become ready after 30 attempts"` with no underlying cause. `getStatusWithRetry` correctly includes `lastErr` — `waitUntilReady` should do the same for debuggability.

None of these block merge. The core change is correct and meaningfully improves test reliability on slower machines.

ghost (approved):
✅ Re-reviewed after new commit `883a1e1c` ("fix: consolidate the funcs").

**Changes since last approval:** Merges the separate `getStatusWithRetry` and `waitUntilReady` into a single generic `waitForStatus(client, attempts, interval, readyFunc)`. Clean refactor, no behavioral regressions.

**Minor observation (non-blocking):**
- `test/docker/setup.go:135-136` — When `ready(status)` returns `false` (HTTP succeeded but predicate not satisfied), no progress log is emitted and the final error has no context about *why* it wasn't ready. Consider adding an else-branch log like `"arkd reachable but not ready yet (attempt %d/%d): initialized=%v unlocked=%v synced=%v"`. Not blocking since this is test infra only.

**Scope:** Test setup only. No protocol-critical code touched. LGTM.

🤖 Arkana

ghost (approved):
✅ Latest commit `1c146aab` adds the status-field logging I suggested — all prior feedback addressed. Still test-infra only, no protocol-critical paths. Good to merge.

🤖 Arkana

Dunsin-cyber:
hi @altafan , is this good to go?

louisinger:
I'm closing it, the current setup.sh is using polling.

PR #190, opened 2026-05-29
VHTLC contract handler and standalone vhtlc package

sekulicd:
Add support for non-derivable contract(that cant be derived without external state) types to the go-sdk contract manager, with VHTLC as the first implementation.                                                                      
                                                                                                                                                                                            
  ### Contract Manager Changes                                                                                                                                                              
  - Add `Derivable() bool` to the `Handler` interface — non-derivable handlers (VHTLC, delegate, covenant) are skipped by `ScanContracts`
  - Add `WithParams(params any)` contract option — passes handler-specific typed parameters through `Manager.NewContract`                                                                   
  - `NewContract` derives the key from the wallet's identity for all types; non-derivable types require `WithParams`                                                                        
  - Skip `ValidateHandler` sanity check (NewContract with random key) for non-derivable handlers                                                                                            
                                                                                                                                                                                            
  ### VHTLC Package (`vhtlc/`)                                                                                                                                                              
  Standalone package for VHTLC script construction, ported from fulmine's `pkg/vhtlc`:                                                                                                      
  - 6 standard spending paths (claim, refund, unilateral exit variants)                                                                                                                     
  - Optional 7th non-interactive claim covenant path with arkade script enforcement                                                                                                         
  - `EnforcePayTo` arkade script builder + `IntrospectorTweakedKey` (inlined, no introspector dependency)                                                                                   
  - `NewVHTLCScriptFromOpts` / `NewVhtlcScript` (recovery from raw tapscript leaves)                                                                                                        
                                                                                                                                                                                            
  ### VHTLC Contract Handler (`contract/handlers/vhtlc/`)                                                                                                                                   
  Stateless handler — all VHTLC parameters stored in `Contract.Params`:                                                                                                                     
  - `Derivable() → false`                                                                                                                                                                   
  - `ContractParams{Opts, Network}` for typed parameter passing                                                                                                                             
  - Supports both interactive (6 leaves) and non-interactive claim (7 leaves)                                                                                                               
  - `OptsFromContract` reconstructs full `vhtlc.Opts` from persisted params                                                                                                                 
  - No role field — `ownerKey` identifies the wallet's participant key                                                                                                                      
                                                                                                                                                                                            
  ### Usage                                                                                                                                                                                 
  ```go                                                                                                                                                                                     
  manager.NewContract(ctx, vhtlcHandler.ContractTypeVHTLC,  
      contract.WithParams(&vhtlcHandler.ContractParams{                                                                                                                                     
          Opts:    opts,                                                                                                                                                                    
          Network: network,                                                                                                                                                                 
      }),                                                                                                                                                                                   
  )                                                         

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added VHTLC support with full tapscript/address handling and optional non-interactive claim flow.
  * Contract creation now accepts handler-specific parameters to support non-derivable contract types.
  * Contract scanning/derivation now skips non-derivable types and derives contracts for derivable handlers.

* **Tests**
  * Updated test harnesses, fixtures, and mocks to align with handler/derivation changes.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/contract/handlers/vhtlc/handler.go
+++ b/contract/handlers/vhtlc/handler.go
@@ -71,7 +71,9 @@ func (h *Handler) NewContract(
 ) (*types.Contract, error) {
 	p, ok := params.(*vhtlc.Opts)
 	if !ok || p == nil {
-		return nil, fmt.Errorf("vhtlc handler requires *vhtlcHandler.ContractParams, got %T", params)
+		return nil, fmt.Errorf(
+			"vhtlc handler requires *vhtlc.Opts, got %T", params,
+		)
 	}
 	return createContract(*p, keyRef, h.network)
 }
--- a/contract/handlers/vhtlc/handler.go
+++ b/contract/handlers/vhtlc/handler.go
@@ -1,0 +1,342 @@
+package vhtlcHandler
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	"strconv"
+	"time"
+
+	arklib "github.com/arkade-os/arkd/pkg/ark-lib"
+	"github.com/arkade-os/arkd/pkg/client-lib/identity"
+	"github.com/arkade-os/go-sdk/contract/handlers"
+	"github.com/arkade-os/go-sdk/types"
+	"github.com/arkade-os/go-sdk/vhtlc"
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/txscript"
+)
+
+// ContractTypeVHTLC is the contract type for VHTLC contracts.
+const ContractTypeVHTLC types.ContractType = "vhtlc"
+
+// Param keys stored in Contract.Params.
+const (
+	paramOwnerKeyID                      = "ownerKeyId"
+	paramOwnerKey                        = "ownerKey"
+	paramSender                          = "sender"
+	paramReceiver                        = "receiver"
+	paramServer                          = "server"
+	paramPreimageHash                    = "preimageHash"
+	paramRefundLocktime                  = "refundLocktime"
+	paramClaimDelayType                  = "claimDelayType"
+	paramClaimDelayValue                 = "claimDelayValue"
+	paramRefundDelayType                 = "refundDelayType"
+	paramRefundDelayValue                = "refundDelayValue"
+	paramRefundWithoutReceiverDelayType  = "refundWithoutReceiverDelayType"
+	paramRefundWithoutReceiverDelayValue = "refundWithoutReceiverDelayValue"
+	paramNICReceiverPkScript             = "nicReceiverPkScript"
+	paramNICIntrospectorPubKey           = "nicIntrospectorPubKey"
+)
+
+const (
+	locktimeTagBlock  = "block"
+	locktimeTagSecond = "second"
+)
+
+// Handler is a stateless contract handler for VHTLC scripts.
+// All VHTLC parameters are stored in Contract.Params, so the handler
+// can rebuild the full tapscript tree from any persisted contract.
+type Handler struct{}
+
+// NewHandler returns a VHTLC contract handler ready to be registered via
+// WithContractHandler(ContractTypeVHTLC, vhtlcHandler.NewHandler()).
+func NewHandler() *Handler { return &Handler{} }
+
+// Derivable returns false — VHTLC contracts require counterparty data
+// (pubkeys, preimage hash, locktimes) and cannot be derived from an HD key alone.
+// Callers must provide WithParams(*ContractParams) when calling Manager.NewContract.
+func (h *Handler) Derivable() bool { return false }
+
+// ContractParams contains the handler-specific parameters for creating a VHTLC
+// contract. Pass via contract.WithParams(&vhtlcHandler.ContractParams{...}).
+type ContractParams struct {
+	Opts    vhtlc.Opts
+	Network arklib.Network
+}
+
+// NewContract builds a VHTLC contract from the caller-provided key and params.
+// params must be *ContractParams; returns an error otherwise.
+func (h *Handler) NewContract(
+	_ context.Context, keyRef identity.KeyRef, params any,
+) (*types.Contract, error) {
+	p, ok := params.(*ContractParams)
+	if !ok || p == nil {
+		return nil, fmt.Errorf("vhtlc handler requires *vhtlcHandler.ContractParams, got %T", params)
+	}
+	return createContract(p.Opts, keyRef, p.Network)
+}
+
+func (h *Handler) GetKeyRef(c types.Contract) (*identity.KeyRef, error) {
+	keyID, err := requireParam(c, paramOwnerKeyID)
+	if err != nil {
+		return nil, err
+	}
+	pubHex, err := requireParam(c, paramOwnerKey)
+	if err != nil {
+		return nil, err
+	}
+	buf, err := hex.DecodeString(pubHex)
+	if err != nil {
+		return nil, fmt.Errorf("vhtlc contract %s: invalid owner key hex: %w", c.Script, err)
+	}
+	pub, err := schnorr.ParsePubKey(buf)
+	if err != nil {
+		return nil, fmt.Errorf("vhtlc contract %s: invalid owner key: %w", c.Script, err)
+	}
+	return &identity.KeyRef{Id: keyID, PubKey: pub}, nil
+}
+
+func (h *Handler) GetKeyRefs(c types.Contract) (map[string]string, error) {
+	keyRef, err := h.GetKeyRef(c)
+	if err != nil {
+		return nil, err
+	}
+	return map[string]string{c.Script: keyRef.Id}, nil
+}
+
+func (h *Handler) GetSignerKey(c types.Contract) (*btcec.PublicKey, error) {
+	return parseCompressedParam(c, paramServer)
+}
+
+// GetExitDelay returns the conservative (longest) exit delay:
+// refundWithoutReceiverDelay. This is always safe regardless of
+// whether the wallet is the sender or receiver.
+func (h *Handler) GetExitDelay(c types.Contract) (*arklib.RelativeLocktime, error) {
+	delay, err := parseRelativeLocktime(
+		c, paramRefundWithoutReceiverDelayType, paramRefundWithoutReceiverDelayValue,
+	)
+	if err != nil {
+		return nil, err
+	}
+	return &delay, nil
+}
+
+func (h *Handler) GetTapscripts(c types.Contract) ([]string, error) {
+	opts, err := OptsFromContract(c)
+	if err != nil {
+		return nil, err
+	}
+	s, err := vhtlc.NewVHTLCScriptFromOpts(opts)
+	if err != nil {
+		return nil, fmt.Errorf("vhtlc contract %s: rebuild script: %w", c.Script, err)
+	}
+	return s.Encode()
+}
+
+// Compile-time check.
+var _ handlers.Handler = (*Handler)(nil)
+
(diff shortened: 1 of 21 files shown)


ghost (changes requested):
## 🔴 PROTOCOL-CRITICAL: VHTLC Review — Requesting Changes

This PR is protocol-critical (VTXO script construction, spending paths, covenant enforcement). **Human sign-off required before merge.**

---

### 🚨 P0 — CRITICAL: Wrong opcode in `EnforcePayTo` (fund loss risk)

**`vhtlc/noninteractive.go:96`**: `opGreaterThanOrEqual` is hardcoded as `0xf4` (244).

The canonical value from the introspector package (`pkg/arkade/opcode.go:207`) is:
```go
OP_GREATERTHANOREQUAL = 0xa2 // 162
```

The fulmine source correctly uses `arkade.OP_GREATERTHANOREQUAL` (0xa2). This PR inlines the wrong value.

**Impact**: The enforcement script hash will be different. The tweaked introspector key will be different. The entire tapscript tree for any VHTLC with `NonInteractiveClaim` enabled will be wrong. **Coins sent to such addresses cannot be claimed via the non-interactive path.** The solver bot's introspector will reject the script because the opcodes don't match what it expects.

**Fix**: Change `opGreaterThanOrEqual = 0xf4` → `opGreaterThanOrEqual = 0xa2`, or better yet, import the constant from the introspector package like fulmine does.

---

### 🚨 P0 — CRITICAL: `IntrospectorTweakedKey` does NOT normalize Y parity

**`vhtlc/noninteractive.go:102-103`**:
```go
// Normalize to even Y before adding.
evenPub, _ := btcec.ParsePubKey(introspectorPubKey.SerializeCompressed())
```

This is a **no-op** — `SerializeCompressed()` preserves the 02/03 prefix, and `ParsePubKey` parses it back with the same parity. The comment says "normalize to even Y" but the code doesn't do it.

The introspector's canonical implementation (`pkg/arkade/tweak.go:24`) does:
```go
pubKey, _ = schnorr.ParsePubKey(schnorr.SerializePubKey(pubKey))
```
This roundtrips through x-only serialization, which **forces even Y parity**.

**Impact**: When `introspectorPubKey` has odd Y, the go-sdk produces a different tweaked key than the introspector. The multisig in the NIC closure will have the wrong second pubkey → same fund-loss scenario as above.

**Fix**: Use `schnorr.ParsePubKey(schnorr.SerializePubKey(introspectorPubKey))` to match the introspector exactly.

---

### 🟡 P1 — No tests ported from fulmine

The fulmine `pkg/vhtlc/` has 460 lines of tests (`vhtlc_test.go`) plus fixture data (`vhtlc_fixtures.json`) covering:
- Address generation from fixtures (deterministic verification)
- Script creation and closure integrity
- Claim/refund witness structure
- `NewVhtlcScript` roundtrip recovery
- Non-interactive claim taptree (7 leaves vs 6)

**None of these were ported.** The `vhtlc/` package in this PR has **zero test coverage**. The `contract/handlers/vhtlc/` handler also has zero tests (342 new lines, 0 test lines). For protocol-critical code that constructs Bitcoin spending conditions, this is unacceptable.

At minimum, port the fulmine fixture tests and add handler-level roundtrip tests for `createContract` → `OptsFromContract` → `GetTapscripts` to verify the tapscript tree is deterministic.

---

### 🟡 P1 — `locktimeTypeName` silently returns empty string for unknown types

**`contract/handlers/vhtlc/handler.go:476`**:
```go
default:
    return ""
```

If an unknown `RelativeLocktimeType` is passed to `createContract`, it stores `""` in the params map. Later, `OptsFromContract` → `parseRelativeLocktime` will fail with "unknown locktime type" on the empty string. This is a silent data corruption → delayed error pattern. Should return an error (or panic) at serialization time, not deserialization.

---

### 🟢 P2 — Minor issues

1. **`vhtlc/vhtlc.go:489` — `Opts()` returns aliased slice**: `v.preimageConditionScript[2 : 2+hash160Len]` returns a slice into internal state. A caller mutating the returned `PreimageHash` corrupts the script. Use `append([]byte(nil), v.preimageConditionScript[2:2+hash160Len]...)` for a defensive copy. (Same issue exists in fulmine — but since this is a new standalone package, fix it now.)

2. **Fragility of inlined opcodes**: Even after fixing the value, hardcoding arkade opcodes as magic numbers is fragile. If the introspector ever bumps an opcode, go-sdk silently diverges. Consider importing from the introspector package or at minimum adding a compile-time assertion test that verifies the constants match.

---

### ✅ What looks good

- **Contract manager changes** are clean: `Derivable()` interface addition, `WithParams` option, skip-scan logic for non-derivable handlers — all correct.
- **Handler interface evolution** is backwards-compatible for external consumers (verified: `solver` and `bancod` only call `GetTapscripts`/`GetKeyRefs`, not `NewContract` directly).
- **Closure construction** (the 6 standard paths) is a faithful port from fulmine.
- **`OptsFromContract` deserialization** is thorough with proper error handling on every field.
- **`GetExitDelay`** returning the conservative `refundWithoutReceiverDelay` is the safe choice.

---

### Summary

Two critical bugs in `noninteractive.go` that would produce wrong tapscript trees and make non-interactive claim paths unspendable. Both are straightforward to fix. Tests must be ported before merge — this is protocol-critical code.

⚠️ **Protocol-critical PR — requires human review even after fixes.**

contract/handlers/vhtlc/handler.go
```diff
@@ -0,0 +1,342 @@
...
+// Derivable returns false — VHTLC contracts require counterparty data
+// (pubkeys, preimage hash, locktimes) and cannot be derived from an HD key alone.
+// Callers must provide WithParams(*ContractParams) when calling Manager.NewContract.
+func (h *Handler) Derivable() bool { return false }
+
+// ContractParams contains the handler-specific parameters for creating a VHTLC
+// contract. Pass via contract.WithParams(&vhtlcHandler.ContractParams{...}).
+type ContractParams struct {
+	Opts    vhtlc.Opts
+	Network arklib.Network
```
altafan: let's follow what's already done for the default handler: the network is passed at contruction time and stored in a `network` private field. This way, we can get rid of `ContractParams` and expect `vhtlc.Opts` to be passed as `param any` in `NewContract`

altafan (changes requested):
We currently miss unit tests in contract/handlers/handler_test.go for the vhtlc handler. Ideally, it should be enough to add it to `modeFixtures` defined at L#53. Probably needed to mock data/svcs to test methods like `NewContract`. The params that are now always hardcoded as `nil`, should likely become a field of the `modeFixture` struct @sekulicd

ghost (commented):
## 🟢 Follow-up Review — P0 Bugs Fixed, Remaining Items

Reviewing commits `a336143`, `1661a34`, `449d53d` (pushed since my last review).

---

### ✅ P0 — Wrong opcode in `EnforcePayTo`: **FIXED**

`vhtlc/noninteractive.go` now imports `github.com/ArkLabsHQ/introspector/pkg/arkade` and uses canonical constants (`arkade.OP_GREATERTHANOREQUAL = 0xa2`, etc.) instead of inlined magic numbers. All custom opcodes removed. Verified the imported package has correct values.

### ✅ P0 — `IntrospectorTweakedKey` Y parity normalization: **FIXED**

`introspectorTweakedKey` now delegates to `arkade.ComputeArkadeScriptPublicKey()` which does the correct `schnorr.ParsePubKey(schnorr.SerializePubKey(pubKey))` roundtrip to force even Y. This matches the introspector's canonical implementation exactly.

### ✅ Handler tests added (altafan's feedback)

VHTLC handler is now in `modeFixtures` with proper coverage: `NewContract` (valid + nil/wrong-type params), `GetKeyRef`, `GetSignerKey`, `GetExitDelay`, `GetTapscripts` — all with valid and invalid cases. Good.

### ✅ `ContractParams` removed — cleaner API

Network moved into `Handler` struct, params simplified to `*vhtlc.Opts` directly. Cleaner.

---

### ⚠️ Still open from previous review

**P1 — `locktimeTypeName` silently returns `""` for unknown types** (`contract/handlers/vhtlc/handler.go:340`)

Still returns empty string on unknown `RelativeLocktimeType`. The callers at lines 173/175/177 store the result directly into `params` — an empty string silently corrupts the persisted contract. `parseRelativeLocktime` then fails later with "unknown locktime type" on deserialization, making it hard to trace back to the root cause. Should return `(string, error)` or at minimum panic on an impossible enum value.

**P1 — No standalone `vhtlc/` package tests**

The handler-level tests provide indirect coverage (good), but the `vhtlc/` package (573 lines of script construction) still has zero test files. Fulmine has 460+ lines of fixture-based tests for this code. At minimum, a single deterministic fixture test (known inputs → known address) would catch any future regressions in tapscript tree construction. This is the kind of code where "it compiles and the handler roundtrips" is not sufficient — you need to verify the exact script bytes match across implementations.

---

### Verdict

The critical fund-loss bugs are fixed. The remaining items are P1 — important but not blocking if there's a plan to add `vhtlc/` package tests before this code hits mainnet. The `locktimeTypeName` issue is a small fix.

Not re-requesting changes since P0s are resolved. Leaving as comment — **human sign-off still required** (protocol-critical).

⚠️ **Protocol-critical PR — requires human review before merge.**

ghost (commented):
## Follow-up Review — Linting Commit `f88c59d2`

Reviewed the new commit. It's **formatting-only** (line-length linting in `handler_test.go` and `handler.go`). No logic changes. ✅ No issues introduced.

### P1 items from my [previous review](https://github.com/arkade-os/go-sdk/pull/190#pullrequestreview-2971277682) still open:

1. **`locktimeTypeName` silently returns `""` for unknown types** (`contract/handlers/vhtlc/handler.go:362-371`) — callers at lines 173/175/177 store the result directly into contract params, silently corrupting persisted data. Should return `(string, error)` or panic on impossible enum.

2. **No standalone `vhtlc/` package tests** — 573 lines of script construction with zero test files. Handler-level tests provide indirect coverage but don't verify exact script bytes match the canonical implementation. Fulmine has 460+ lines of fixture-based tests for this code.

Not blocking merge on these (P0s were fixed), but flagging again for visibility.

⚠️ **Protocol-critical PR — requires human review before merge.**

contract/handlers/vhtlc/handler.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	return &identity.KeyRef{Id: keyID, PubKey: pub}, nil
+}
+
+func (h *Handler) GetKeyRefs(c types.Contract) (map[string]string, error) {
+	keyRef, err := h.GetKeyRef(c)
+	if err != nil {
+		return nil, err
+	}
+	return map[string]string{c.Script: keyRef.Id}, nil
```
altafan, line 106: we need to add here also all possible checkpoint scripts all committing to the same key id `keyRef.Id` depending if we are sender or receiver:
- claim path + checkpoint exit path if we're receiver
- refund alone path + checkpoint exit path if we're sender
- collab refund path + checkpoint exit path if we're sender

contract/handlers/vhtlc/handler.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	s, err := vhtlc.NewVHTLCScriptFromOpts(opts)
+	if err != nil {
+		return nil, fmt.Errorf("vhtlc contract %s: rebuild script: %w", c.Script, err)
+	}
+	return s.Encode()
+}
+
+// Compile-time check.
+var _ handlers.Handler = (*Handler)(nil)
```
altafan, line 139: drop

contract/manager.go
```diff
@@ -325,9 +326,9 @@ scan:
 	return nil
 }
 
-func (m *contractManager) newContract(
+func (m *contractManager) deriveContract(
```
altafan, line 329: revert, rename is not necessary

PR #191, opened 2026-05-29
Port swap package and add VHTLC e2e tests

sekulicd:
Port the swap and boltz packages from fulmine and add comprehensive e2e tests including non-interactive claim.

  Swap Package (swap/, swap/boltz/)

  Ported from fulmine's pkg/swap and pkg/boltz (offline-ln-receive branch):
  - Submarine swaps (Ark → Lightning)
  - Reverse swaps (Lightning → Ark)
  - Chain swaps (Ark ↔ BTC on-chain) with cooperative and unilateral refund
  - MuSig2 batch session handling for VHTLC settlement
  - Fix data races in ChainSwap — added sync.RWMutex with thread-safe getters

  Wallet Interface

  - Add GetConfigData to Wallet interface (needed by swap handler)
  - Add ...SendOption passthrough to SendOffChain (enables WithExtraPacket, WithTxOutsTaprootTree)
  - Remove btcec/v2 replace directive (enables lnd v0.19.1-beta compatibility)

  E2E Tests

  - 6 VHTLC tests: claim (direct + settlement), refund, delegate refund, outpoint selection, oldest VTXO
  - 13 swap tests: submarine, reverse, chain swaps, cooperative/unilateral refund, event injection, concurrent swaps
  - 1 non-interactive claim test: go-sdk wallet funds VHTLC with encrypted preimage extension → solver (bancod) auto-claims → funds arrive at receiver

  Test Infrastructure (test/infra/)

  - Docker compose with all services: arkd, emulator, solver, fulmine, boltz, mock-boltz, LND, CLN
  - All services build from GitHub master (no local repo dependency)
  - Setup script provisions arkd, solver, LN channels, and fulmine wallets

## Summary by CodeRabbit

* **New Features**
  * VHTLC support (including non‑interactive claim), HTLC support, BOLT12 offer/invoice decoding, full swap orchestration (submarine/reverse/chain) with Boltz integration and cooperative MuSig2 signing; SDK: config retrieval and extended off‑chain send options; chain‑swap resume/monitoring and on‑chain claim/refund flows.

* **Tests**
  * Large expansion of unit and end‑to‑end swap/VHTLC tests and new regtest E2E suites.

* **Chores**
  * Test/integration workflow moved to a new infra-based regtest setup with provisioning scripts and Docker compose.


ghost (changes requested):
# 🔍 Arkana Code Review — Protocol-Critical

This PR adds the VHTLC contract handler, a full Boltz-based swap subsystem, and significant SDK/contract API changes. Thorough review across all subsystems. **This is protocol-critical code — multiple findings involve direct fund loss risk or private key compromise.**

---

## 🔴 CRITICAL — Must fix before merge

### C1. `ParseEvent` returns `SwapCreated` (zero-value) for unknown status strings
**`swap/boltz/status.go:42-44`**

```go
func ParseEvent(event string) SwapUpdateEvent {
    return swapUpdateEventStrings[event]
}
```
If Boltz sends an unrecognized status string, this returns `0` == `SwapCreated`. Combined with **C2** below, this means a malformed WebSocket message can trigger `HandleSwapCreated`, which in the Ark-to-BTC flow sends the user's VTXO to the lockup address. **This is a fund loss vector.**

Fix: Return a sentinel "unknown" event and handle it explicitly in the monitor.

### C2. WebSocket decode failure sends zero-value update → triggers `SwapCreated`
**`swap/boltz/ws.go:101-104`**

```go
var update SwapUpdate
if err := mapstructure.Decode(arg, &update); err != nil {
}
boltz.Updates <- update
```

If decode fails, a zero-valued `SwapUpdate` (empty status → `ParseEvent("")` → `SwapCreated`) is sent to the consumer. Combined with C1, this is a direct fund loss path.

Fix: Skip sending on decode error; log the error.

### C3. MuSig2 nonce stored in struct but never zeroed after use — reuse leaks private key
**`swap/musig.go:57-66`**

`GenerateNonce()` stores `c.ourNonces` and `OurPartialSign()` uses `c.ourNonces.SecNonce`. If `OurPartialSign` is called twice (e.g., retry logic), the same `SecNonce` is reused — this is a **catastrophic MuSig2 vulnerability** that leaks the full private key to anyone who sees both partial signatures.

Fix: Zero `c.ourNonces.SecNonce` after first use in `OurPartialSign`, and return an error if it's already zeroed.

### C4. `IntrospectorTweakedKey` does NOT normalize to even Y (despite comment claiming it does)
**`vhtlc/noninteractive.go:100-102`**

```go
// Normalize to even Y before adding.
evenPub, _ := btcec.ParsePubKey(introspectorPubKey.SerializeCompressed())
```

`SerializeCompressed()` preserves original Y parity (`0x02` or `0x03`). `ParsePubKey` reconstructs the same point. **This is a no-op, not a normalization.** If the introspector has odd Y, the solver will compute a different tweaked key than what's embedded in the tapscript → spending path permanently unspendable → **funds locked forever** for ~50% of introspector keys.

Fix: Actually negate the key when `serialized[0] == 0x03`.

### C5. `NewVhtlcScript` cannot reconstruct scripts with NonInteractiveClaim (7th leaf)
**`vhtlc/vhtlc.go:107-192`**

Reconstruction path only handles 6 leaf scripts. If a VHTLC was created with `NonInteractiveClaim`, the reconstructed tap tree has a different root hash → merkle proofs fail → **spending via any script path fails → funds locked.**

Fix: Accept optional 7th leaf or NIC opts in `NewVhtlcScript`.

### C6. `server` key can be nil without error in `NewVhtlcScript`
**`vhtlc/vhtlc.go:161-167`**

```go
var server *btcec.PublicKey
for _, pk := range claimClosure.PubKeys {
    if pk.IsEqual(receiver) { continue }
    server = pk
}
```

If both keys equal `receiver` (malformed script), `server` remains nil → nil-pointer dereference downstream in `Address()` or `GetSignerKey()`.

Fix: Return an error if `server` is nil after the loop.

### C7. Race condition: ChainSwap status read without lock in monitor loop
**`swap/chainswap_monitor.go:114-120`**

Direct field access to `chainSwapState.Swap.Status` without holding the mutex, while event handlers modify `Status` under lock. Data race on non-x86 architectures → monitor could miss terminal state and loop forever, or read a torn value.

Fix: Use `chainSwapState.Swap.GetStatus()` which takes `RLock()`.

---

## 🟠 HIGH — Strong recommendation to fix

### H1. No server lockup amount validation in Ark-to-BTC flow
**`swap/chainswap_arkt_btc_handler.go:115-121`**

When Boltz locks BTC, the handler claims without verifying the locked amount matches `swapResp.ClaimDetails.Amount`. A malicious/buggy Boltz could lock less BTC than promised → user loses the difference.

### H2. Fee estimation calls `computeVSize` on unsigned transaction
**`swap/btc_util.go:53, 61-65`**

Witness data is empty when `computeVSize` runs, so `totalSize == baseSize`, producing non-segwit weight. The `+100` sats padding is arbitrary and unreliable. Real witness data (sig + preimage + script) could exceed 100 bytes easily.

### H3. WebSocket reconnect loop can spin forever
**`swap/boltz/ws.go:114-127`**

No maximum retry count and no context cancellation check. If the server is permanently down, this goroutine leaks resources forever.

### H4. `GetExitDelay` assumes `refundWithoutReceiverDelay` is always longest
**`contract/handlers/vhtlc/handler.go:112-122`**

No validation that `UnilateralRefundWithoutReceiverDelay >= UnilateralClaimDelay`. If a caller sets `UnilateralClaimDelay` longer, `GetExitDelay` returns too-short a delay → premature exit attempt.

### H5. Explorer client bypasses configured HTTP client
**`swap/explorer.go:118, 147`**

`GetCurrentBlockHeight` and `GetTransactionStatus` use `http.Get()` instead of `e.client.Get()`, bypassing the 10-second timeout → calls can hang indefinitely.

### H6. Unconditional quote acceptance with no bounds checking
**`swap/chainswap_arkt_btc_handler.go:146-155`**

When a quote is fetched after lockup failure, it's accepted without validating that `quote.Amount` is reasonable. Boltz could return drastically reduced amounts.

### H7. Context from caller passed to swap goroutine — cancellation kills mid-flight swap
**`swap/chainswap.go:335-346`**

In `ChainSwapArkToBtc`, the goroutine uses the caller's `ctx`. If the caller cancels, the swap fails mid-flight with funds already locked. `ChainSwapBtcToArk` correctly uses `context.Background()` — inconsistency here is dangerous.

---

## 🟡 MEDIUM

### M1. `Opts()` roundtrip loses NonInteractiveClaim data on reconstructed scripts
**`vhtlc/vhtlc.go:295-308`** — Roundtrip through `Opts()` → `NewVHTLCScriptFromOpts()` produces different script (6 vs 7 leaves).

### M2. No validation that sender ≠ receiver in VHTLC opts
**`vhtlc/opts.go:31-66`** — Equal keys → server extraction fails (see C6).

### M3. No HTLC delay ordering validation
**`vhtlc/opts.go:31-66`** — Claim delay should be shorter than refund delay. Without this check, sender can refund before receiver claims.

### M4. `parsePubkey` returns nil for empty string without error
**`swap/utils.go:337-339`** — Used in `chainswap_monitor.go:30` where result goes to `NewArkToBtcHandler` without nil check.

### M5. Updates channel is unbuffered
**`swap/boltz/ws.go:41`** — Slow consumer blocks read goroutine → pong messages missed → connection timeout → reconnect → missed updates.

### M6. `ResumeChainSwap` doesn't set `IsArkToBtc` field
**`swap/chainswap_resume.go:94`** — Defaults to `false`, wrong for resumed Ark-to-BTC swaps.

### M7. `locktimeTypeName` returns empty string for unknown types
**`contract/handlers/vhtlc/handler.go:332-340`** — Silent data corruption on persist.

### M8. BTC-to-Ark handler accesses ChainSwap fields without lock during refund
**`swap/chainswap_btc_ark_handler.go:155-160`** — Direct field access to `Amount`, `UserLockTxid`, `SwapRespJson` without mutex.

---

## 🟢 LOW / Notes

- `vhtlc/noninteractive.go:92`: `PrivKeyFromBytes` silently reduces mod N — document this behavior.
- `vhtlc/noninteractive.go:61`: `EnforcePayTo` doesn't check script type before slicing.
- `swap/bolt12.go:49-51`: Integer division truncates millisatoshis → should round up.
- `swap/explorer.go:56-65`: `BroadcastTransaction` body-reading structure is fragile.
- `contract/handlers/vhtlc/handler.go:96-103`: Owner key stored as x-only while others are compressed — asymmetry may confuse.
- `send.go:57`: Caller-supplied `SendOption`s can override SDK internal options (change address) — document this.
- API breaks to `Handler` and `Wallet` interfaces are backward-compatible in practice (no downstream implementors found in `/root/arkana/repos/`).

---

## Verdict

**Requesting changes.** The critical findings (C1-C7) involve direct fund loss risk, key compromise risk, and data races. The VHTLC even-Y normalization bug (C4) and the `ParseEvent` zero-value bug (C1+C2) are the most dangerous — both can silently lose funds in production.

⚠️ **Protocol-critical flag**: This PR touches VTXO handling, tapscript construction, transaction signing (MuSig2), and exit paths. **Human review required** even after fixes.

🤖 *Reviewed by Arkana*

ghost (changes requested):
# 🔍 Arkana Code Review — Iterative (new commits since Jun 2)

Reviewing commits `a3361435..faefcfb7` (7 new commits). The private key removal from swap (`faefcfb7`) and new HTLC contract handler (`abf2392c`) are the major changes.

---

## ✅ What's improved

**Private key removal from swap** — Well executed. `SwapHandler` no longer stores `*btcec.PrivateKey`. All signing delegated to `identity.KeyedTaprootSigner` via contract manager. The `keyedMuSig2Session` in `identity/taproot_signer.go` is a faithful 1:1 reimplementation of the old `MuSigContext`. Architecture is cleaner.

**C2 fixed** — WebSocket decode failure now `continue`s instead of pushing zero-value update. Good.

**New HTLC handler** — `contract/handlers/htlc/handler.go` correctly returns `Derivable() = false`. Script validation is thorough. Address verification before storage is correct.

---

## 🔴 CRITICAL — Still open from previous review

### C1. `ParseEvent` still returns `SwapCreated` for unknown status (NOT FIXED)
**`swap/boltz/status.go:~49`**

Go map zero-value returns `SwapCreated` for any unrecognized event string. Combined with the monitor's switch statement, an unknown Boltz status triggers the `SwapCreated` handler — fund loss vector. This was flagged Jun 2.

### C3. MuSig2 nonce never zeroed after signing (NOT FIXED — now in new location)
**`identity/taproot_signer.go:~130-140`**

The nonce moved from `swap/musig.go` to `identity/taproot_signer.go` but the bug persists. After `musig2.Sign(s.ourNonces.SecNonce, ...)`, `s.ourNonces` is never nilled. If `PartialSign` is called twice with different signing data, this is catastrophic MuSig2 nonce reuse → **full private key extraction**.

Fix: Add `s.ourNonces = nil` after the `musig2.Sign` call. Return error if `s.ourNonces == nil` at entry.

### C4. `IntrospectorTweakedKey` even-Y normalization is still a no-op (NOT FIXED)
**`vhtlc/noninteractive.go:72-77`**

No even-Y check or negation. ~50% of introspector keys will compute a wrong tweaked key → funds permanently locked.

### C5. `NewVhtlcScript` still cannot reconstruct 7-leaf scripts (NOT FIXED)
**`vhtlc/vhtlc.go:107-192`**

VHTLCs with `NonInteractiveClaimOpts` still can't round-trip. Different tap root → script-path spending fails.

### C6. `server` key nil after loop still unchecked (NOT FIXED)
**`vhtlc/vhtlc.go:161-168`**

### C7. Race condition in monitor loop still present (NOT FIXED)
**`swap/chainswap_monitor.go:163-168`**

Direct `chainSwapState.Swap.Status` reads without mutex while handlers write under lock. Use `GetStatus()`.

---

## 🔴 NEW CRITICAL — Introduced in this update

### N1. Private key lifetime in MuSig2 sessions
**`identity/taproot_signer.go:74-79`**

`NewMuSig2Session()` derives a private key under `RLock()`, stores it in the returned `keyedMuSig2Session`. After the lock releases, the session holds a live `*btcec.PrivateKey` indefinitely. If `Identity.Lock()` is called (zeroing the mnemonic), outstanding sessions still hold usable private keys. The key is never zeroed after `PartialSign`.

Fix: Zero `s.key` after `PartialSign`. Document session lifecycle constraints.

---

## 🟠 HIGH — Still open

### H5. Explorer still uses bare `http.Get()` (NOT FIXED)
**`swap/explorer.go:~107, ~125, ~158`**

Three methods bypass `e.client` with its 10s timeout. Calls can hang indefinitely.

### H7. `ChainSwapArkToBtc` still uses caller's context (PARTIALLY FIXED)
**`swap/chainswap.go:~360`**

`BtcToArk` correctly uses `context.Background()`. `ArkToBtc` still passes caller ctx → cancellation kills mid-flight swap with funds locked.

### H8 (NEW). `newLocalHTLCKey` missing nil validation
**`swap/htlc_contracts.go:14-19`**

Unlike `newLocalVHTLCKey` (which checks `keyRef == nil || keyRef.PubKey == nil || keyRef.Id == ""`), `newLocalHTLCKey` does zero validation. Nil `PubKey` → panic in `schnorr.SerializePubKey`.

### H9 (NEW). MuSig2 key ordering must match Boltz protocol
**`contract/handlers/htlc/handler.go:173-176`**

Keys aggregated as `[Server, Owner]` with sorting disabled. If Boltz uses `[Owner, Server]`, the aggregate key differs → funds sent to wrong address → unspendable. This ordering must be verified against the Boltz protocol spec.

---

## 🟡 MEDIUM — Still open + new

### M6. `ResumeChainSwap` doesn't set `IsArkToBtc` (NOT FIXED)
**`swap/chainswap_resume.go:~113-128`**

`arkToBtc` computed at line 80 but never assigned to struct. Resumed Ark-to-BTC swaps default to BTC-to-Ark behavior.

### M8. BTC-to-Ark handler field access without lock (NOT FIXED)
**`swap/chainswap_btc_ark_handler.go:155-160`**

### M9 (NEW). HTLC handler doesn't enforce owner key leaf role
**`contract/handlers/htlc/handler.go:170-171`**

Owner key accepted in either claim or refund leaf. Malicious server could put user's key in wrong leaf, complicating fund recovery.

### M10 (NEW). `ensureLocalHTLCContract` swallows build errors
**`swap/htlc_contracts.go:56-58`**

`if err != nil || built.Address != expectedAddress { continue }` — build errors silently discarded. Makes debugging key-mismatch failures hard.

### M11 (NEW). VHTLC `findLocalVHTLCKey` matches sender OR receiver without role check
**`swap/vhtlc_contracts.go:86-92`**

No validation that the matched role is appropriate for the operation being performed.

---

## 🔗 Cross-repo impact

**fulmine will break** when bumping go-sdk:
- `VHTLCHandler` at `fulmine/pkg/swap/vhtlc_handler.go:75` doesn't implement new `Derivable() bool` method
- `NewContract` signature changed (added `params any` parameter)
- Consider coordinating a fulmine PR

---

## Verdict

**Requesting changes (again).** 6 of 7 original critical findings remain unfixed (C1, C3-C7). One new critical finding (N1) introduced. The private key removal architecture is solid, but the same nonce-reuse bug (C3) migrated to the new location.

Priority order for fixes:
1. **C3/N1** — Nonce zeroing + key zeroing in `taproot_signer.go` (money-losing MuSig2 bug)
2. **C4** — Even-Y normalization (funds permanently locked)
3. **C1** — ParseEvent zero-value (fund loss vector)
4. **C7** — Data race in monitor (can cause infinite loop)
5. **M6** — IsArkToBtc not set on resume (wrong swap direction behavior)

⚠️ **Protocol-critical flag**: This PR touches transaction signing (MuSig2), tapscript construction, VTXO handling, and exit paths. **Human review required** even after all fixes.

🤖 *Reviewed by Arkana*

ghost (changes requested):
# 🔍 Arkana Code Review — Iterative (new commits since Jun 3)

Focus: commits `e6caf496..d0d04b4d` (Jun 5 – Jun 19). ~4000 lines changed across 59 files.

---

## 🚨 CRITICAL — Protocol-Critical / Loss-of-Funds Risk

### 1. Preimage revealed on unconfirmed (mempool) BTC lockup
`swap/chainswap_arkt_btc_handler.go:82-85` → `HandleServerLockedMempool` calls `handleArkToBtcServerLocked`, which claims BTC and **sends the preimage to Boltz** via `SubmitChainSwapClaim` (line ~330).

The problem: at `TransactionServerMempool`, Boltz's BTC lockup is **only in mempool** — not confirmed. Boltz can double-spend the lockup transaction after receiving the preimage, then claim the user's Ark VHTLC using the revealed preimage. User loses both sides.

`HandleServerLocked` (confirmed state, line 88) is a no-op `return nil` — meaning the claim ONLY happens on mempool, never on confirmation. This is backwards for safety.

**Fix:** Move the claim logic to `HandleServerLocked` (confirmed). `HandleServerLockedMempool` should only log. Alternatively, build the claim tx on mempool but only reveal the preimage after confirmation.

### 2. ARK→BTC monitoring goroutine uses caller context
`swap/chainswap.go:536` passes `ctx` (the caller's context) to `monitorAndClaimArkToBtcSwap`. When `ChainSwapArkToBtc` returns and the caller's context is cancelled, the monitoring goroutine dies — potentially abandoning a swap with real BTC locked.

Compare with `swap/chainswap.go:708` (BTC→ARK) which correctly uses `context.Background()`. This is an asymmetric bug.

### 3. Auto-quote acceptance without bounds checking
`swap/chainswap_arkt_btc_handler.go:193-211`: On `TransactionLockupFailed`, the code fetches a quote from Boltz and immediately accepts it with no validation of `quote.Amount` or `quote.OnchainAmount` against the original swap terms. A malicious Boltz could return an exploitative quote (e.g., claiming 90% as fees). Same pattern at `swap/chainswap_btc_ark_handler.go:173-197`.

**Fix:** Validate that `quote.Amount` is within an acceptable tolerance of the original swap amount before accepting.

---

## ⚠️ HIGH — Correctness / Reliability

### 4. No WebSocket reconnection for chain swap monitoring
`swap/chainswap_monitor.go:108-110`: When the WebSocket channel closes, the chain swap is immediately **permanently failed** via `chainSwapState.Swap.Fail("websocket closed")`. For chain swaps involving real BTC, a transient network glitch permanently kills the swap.

Compare with submarine swap monitoring at `swap/swap.go:901-909` which has reconnection logic. Chain swaps need this too.

### 5. `parsePubkey("")` silently returns `(nil, nil)`
`swap/utils.go:381-384`: Empty string returns nil key with no error. Callers like `btc_htlc_contracts.go:24` (`newHTLCOpts`) and chain swap lockup validation pass the result into MuSig2 or script construction. If Boltz API returns an empty `serverPublicKey`, downstream code gets a nil key instead of an error.

**Fix:** Return `fmt.Errorf("empty public key")` for empty input.

### 6. `RefundBtcToArkSwap` blocks for up to 2 hours, ignores context
`swap/chainswap.go:958-982`: A 2-hour polling loop with 30-second intervals that never checks `ctx.Done()`. The caller has no way to cancel. The function signature accepts a `ctx` but ignores it in the loop.

### 7. `combineTapscripts` panics on empty TaprootScriptSpendSig
`swap/utils.go:339-343`: Indexes `boltzIn.TaprootScriptSpendSig[0]` without checking if the slice is non-empty. If any PSBT input from Boltz lacks a tapscript sig, this panics. Compare with `swap.go:545-546` which properly checks `len(...) == 0`.

---

## 🟡 MEDIUM — Robustness / Defense in Depth

### 8. `checkpointExitScript` swallows hex decode error
`swap/utils.go:32-34`: `hex.DecodeString` error discarded via `_`. Invalid `CheckpointTapscript` config results in nil byte slice propagated silently.

### 9. `Lock()` doesn't wait for background goroutines
`init.go:180-215`: Unlike `Stop()` which calls `waitForBackground(5s)`, `Lock()` cancels the context and immediately nils out `contractManager`, `stopCtx`, etc. Still-running background goroutines (e.g., `periodicRefreshDb` → `refreshDb` → `contractManager.GetContracts()`) can dereference nil pointers. Goroutines from a previous Unlock cycle may overlap with the next.

### 10. Potential uint64 underflow in amount calculations
Multiple locations: `wallet.go:1598-1601` (`amount -= v.Amount`), `wallet.go:1729-1737` (`inAmount - outAmount`), `wallet.go:2096` (`spentAmount -= rcv.Amount`). If server returns outputs exceeding inputs (buggy or malicious), uint64 wraps to astronomical values. While this doesn't directly lose funds, it corrupts transaction history and could confuse the user into making bad decisions.

### 11. `saveSendTransaction` swallows `AddTransactions` error
`wallet.go:2170-2173`: Error is logged with `Warnf` and the function returns `nil`. Caller sees success even though the transaction record was not persisted. For a financial application, this should propagate the error.

---

## 📝 LOW — Code Quality / Minor

### 12. Fee estimation magic constant
`swap/btc_util.go:72`: `feeAmount := uint64(math.Ceil(float64(vbytes)*feeRate) + 100)` — unexplained +100 sat. The vbytes is also computed on the unsigned tx (underestimates final size with witness). Should be documented or computed more precisely.

### 13. `listenForArkTxs` hot loop on closed channel
`wallet.go:934-936`: When `eventChan` is closed, `ok == false` triggers `continue`, which immediately re-enters the `select` and fires the closed channel case again in a tight CPU spin. Should `return` instead of `continue`.

### 14. `parseLocktime` threshold undocumented
`swap/utils.go:322-327`: The `>= 512` threshold distinguishes block-based vs. time-based locktimes. This is the BIP-68 relative locktime threshold but deserves a comment explaining the convention, as it's a subtle protocol detail.

---

## ✅ Positive Changes Since Last Review

- **txHandler pattern** (`tx_handler.go`): Well-designed priority queue serializing spend/batch operations. Prevents double-spend of VTXOs. No deadlock risk identified.
- **Test matrix** (`.github/workflows/integration.yaml`, `Makefile`): Good parallelization of CI with `max-parallel: 6`.
- **Rename `arkClient` → `arkWallet`** throughout `batch_handler.go`: Clearer naming.
- **Removed `publicKey` from `refundBatchSessionHandler`**: Reduces surface area.
- **`notifyTracked` pattern** (subscribe-before-submit): Correctly implemented across `send.go`, `batch_session.go`, and `asset.go`.

---

**⛔ Requesting changes.** Items #1 (preimage on mempool), #2 (caller context in monitoring), and #3 (unbounded quote acceptance) are loss-of-funds vectors that must be addressed before merge.

**🔴 PROTOCOL-CRITICAL** — This PR touches VTXO handling, transaction signing, forfeit/refund paths, and swap settlement. Human sign-off required even after fixes.

🤖 *Reviewed by Arkana*

ghost (changes requested):
# 🔍 Arkana Code Review — Iterative (new commits since Jun 3)

Focus: commits `e6caf496..d0d04b4d` (Jun 5 – Jun 19). ~4000 lines changed across 59 files.

---

## 🚨 CRITICAL — Protocol-Critical / Loss-of-Funds Risk

### 1. Preimage revealed on unconfirmed (mempool) BTC lockup
`swap/chainswap_arkt_btc_handler.go:82-85` → `HandleServerLockedMempool` calls `handleArkToBtcServerLocked`, which claims BTC and **sends the preimage to Boltz** via `SubmitChainSwapClaim` (line ~330).

The problem: at `TransactionServerMempool`, Boltz's BTC lockup is **only in mempool** — not confirmed. Boltz can double-spend the lockup transaction after receiving the preimage, then claim the user's Ark VHTLC using the revealed preimage. User loses both sides.

`HandleServerLocked` (confirmed state, line 88) is a no-op `return nil` — meaning the claim ONLY happens on mempool, never on confirmation. This is backwards for safety.

**Fix:** Move the claim logic to `HandleServerLocked` (confirmed). `HandleServerLockedMempool` should only log. Alternatively, build the claim tx on mempool but only reveal the preimage after confirmation.

### 2. ARK→BTC monitoring goroutine uses caller context
`swap/chainswap.go:536` passes `ctx` (the caller's context) to `monitorAndClaimArkToBtcSwap`. When `ChainSwapArkToBtc` returns and the caller's context is cancelled, the monitoring goroutine dies — potentially abandoning a swap with real BTC locked.

Compare with `swap/chainswap.go:708` (BTC→ARK) which correctly uses `context.Background()`. This is an asymmetric bug.

### 3. Auto-quote acceptance without bounds checking
`swap/chainswap_arkt_btc_handler.go:193-211`: On `TransactionLockupFailed`, the code fetches a quote from Boltz and immediately accepts it with no validation of `quote.Amount` or `quote.OnchainAmount` against the original swap terms. A malicious Boltz could return an exploitative quote (e.g., claiming 90% as fees). Same pattern at `swap/chainswap_btc_ark_handler.go:173-197`.

**Fix:** Validate that `quote.Amount` is within an acceptable tolerance of the original swap amount before accepting.

---

## ⚠️ HIGH — Correctness / Reliability

### 4. No WebSocket reconnection for chain swap monitoring
`swap/chainswap_monitor.go:108-110`: When the WebSocket channel closes, the chain swap is immediately **permanently failed** via `chainSwapState.Swap.Fail("websocket closed")`. For chain swaps involving real BTC, a transient network glitch permanently kills the swap.

Compare with submarine swap monitoring at `swap/swap.go:901-909` which has reconnection logic. Chain swaps need this too.

### 5. `parsePubkey("")` silently returns `(nil, nil)`
`swap/utils.go:381-384`: Empty string returns nil key with no error. Callers like `btc_htlc_contracts.go:24` (`newHTLCOpts`) and chain swap lockup validation pass the result into MuSig2 or script construction. If Boltz API returns an empty `serverPublicKey`, downstream code gets a nil key instead of an error.

**Fix:** Return `fmt.Errorf("empty public key")` for empty input.

### 6. `RefundBtcToArkSwap` blocks for up to 2 hours, ignores context
`swap/chainswap.go:958-982`: A 2-hour polling loop with 30-second intervals that never checks `ctx.Done()`. The caller has no way to cancel. The function signature accepts a `ctx` but ignores it in the loop.

### 7. `combineTapscripts` panics on empty TaprootScriptSpendSig
`swap/utils.go:339-343`: Indexes `boltzIn.TaprootScriptSpendSig[0]` without checking if the slice is non-empty. If any PSBT input from Boltz lacks a tapscript sig, this panics. Compare with `swap.go:545-546` which properly checks `len(...) == 0`.

---

## 🟡 MEDIUM — Robustness / Defense in Depth

### 8. `checkpointExitScript` swallows hex decode error
`swap/utils.go:32-34`: `hex.DecodeString` error discarded via `_`. Invalid `CheckpointTapscript` config results in nil byte slice propagated silently.

### 9. `Lock()` doesn't wait for background goroutines
`init.go:180-215`: Unlike `Stop()` which calls `waitForBackground(5s)`, `Lock()` cancels the context and immediately nils out `contractManager`, `stopCtx`, etc. Still-running background goroutines (e.g., `periodicRefreshDb` → `refreshDb` → `contractManager.GetContracts()`) can dereference nil pointers. Goroutines from a previous Unlock cycle may overlap with the next.

### 10. Potential uint64 underflow in amount calculations
Multiple locations: `wallet.go:1598-1601` (`amount -= v.Amount`), `wallet.go:1729-1737` (`inAmount - outAmount`), `wallet.go:2096` (`spentAmount -= rcv.Amount`). If server returns outputs exceeding inputs (buggy or malicious), uint64 wraps to astronomical values. While this doesn't directly lose funds, it corrupts transaction history and could confuse the user into making bad decisions.

### 11. `saveSendTransaction` swallows `AddTransactions` error
`wallet.go:2170-2173`: Error is logged with `Warnf` and the function returns `nil`. Caller sees success even though the transaction record was not persisted. For a financial application, this should propagate the error.

---

## 📝 LOW — Code Quality / Minor

### 12. Fee estimation magic constant
`swap/btc_util.go:72`: `feeAmount := uint64(math.Ceil(float64(vbytes)*feeRate) + 100)` — unexplained +100 sat. The vbytes is also computed on the unsigned tx (underestimates final size with witness). Should be documented or computed more precisely.

### 13. `listenForArkTxs` hot loop on closed channel
`wallet.go:934-936`: When `eventChan` is closed, `ok == false` triggers `continue`, which immediately re-enters the `select` and fires the closed channel case again in a tight CPU spin. Should `return` instead of `continue`.

### 14. `parseLocktime` threshold undocumented
`swap/utils.go:322-327`: The `>= 512` threshold distinguishes block-based vs. time-based locktimes. This is the BIP-68 relative locktime threshold but deserves a comment explaining the convention, as it's a subtle protocol detail.

---

## ✅ Positive Changes Since Last Review

- **txHandler pattern** (`tx_handler.go`): Well-designed priority queue serializing spend/batch operations. Prevents double-spend of VTXOs. No deadlock risk identified.
- **Test matrix** (`.github/workflows/integration.yaml`, `Makefile`): Good parallelization of CI with `max-parallel: 6`.
- **Rename `arkClient` → `arkWallet`** throughout `batch_handler.go`: Clearer naming.
- **Removed `publicKey` from `refundBatchSessionHandler`**: Reduces surface area.
- **`notifyTracked` pattern** (subscribe-before-submit): Correctly implemented across `send.go`, `batch_session.go`, and `asset.go`.

---

**⛔ Requesting changes.** Items #1 (preimage on mempool), #2 (caller context in monitoring), and #3 (unbounded quote acceptance) are loss-of-funds vectors that must be addressed before merge.

**🔴 PROTOCOL-CRITICAL** — This PR touches VTXO handling, transaction signing, forfeit/refund paths, and swap settlement. Human sign-off required even after fixes.

🤖 *Reviewed by Arkana*

ghost (changes requested):
# 🔍 Arkana Code Review — Iterative (commit `914bdd11ac`)

New commit since last review: formatting fixes (line wrapping in `htlc/handler.go`, `contract/manager.go`) and Dockerfile infra update (Go 1.26.3→1.26.4, repo URL fix to `arkade-os`). **No functional changes.**

The formatting and infra changes look fine. ✅

---

## ⛔ Outstanding Critical Issues (unchanged)

All critical findings from my previous review remain unaddressed. Reiterating the top 3 that block approval:

**1. Preimage revealed on unconfirmed BTC lockup — loss of funds**
`swap/chainswap_arkt_btc_handler.go:79-88`: `HandleServerLockedMempool` claims BTC and reveals preimage while Boltz's lockup is only in mempool. `HandleServerLocked` (confirmed) is a no-op. Boltz can double-spend the unconfirmed lockup after receiving the preimage, then claim user's Ark VHTLC. User loses both sides.

**2. Monitoring goroutine uses caller context — swap abandonment**
`swap/chainswap.go:536`: `monitorAndClaimArkToBtcSwap` receives caller's `ctx`. When caller returns and context cancels, the goroutine dies, abandoning a swap with real BTC locked. BTC→ARK path at line 707 correctly uses `context.Background()`.

**3. Unbounded quote acceptance — fee exploitation**
`swap/chainswap_arkt_btc_handler.go:193-211` and `swap/chainswap_btc_ark_handler.go:173-197`: Boltz quote accepted without validating amount against original swap terms. Malicious or buggy Boltz could extract excessive fees.

Please address these before merge. This is protocol-critical code.

🤖 *Reviewed by Arkana*

sekulicd:
@arkanaai can you do another deep review

sekulicd:
@coderabbitai can you do another deep review.

contract/handlers/htlc/handler.go
```diff
@@ -0,0 +1,327 @@
...
+
+func tapLeafHash(script []byte) [32]byte {
+	var b bytes.Buffer
+	b.WriteByte(supportedLeafVersion)
+	_ = wire.WriteVarInt(&b, 0, uint64(len(script)))
+	b.Write(script)
+	sum := chainhash.TaggedHash(chainhash.TagTapLeaf, b.Bytes())
+
+	return *sum
+}
```
louisinger: could we replace by `txscript.AssembleTaprootScriptTree` ?

contract/handlers/htlc/handler.go
```diff
@@ -0,0 +1,327 @@
...
+
+// Opts are the parameters needed to create a BTC HTLC lockup contract.
+// Server is the Boltz/server MuSig key. The wallet-owned key is supplied by
+// the contract manager as keyRef and must appear in either the claim or refund
+// leaf script.
+type Opts struct {
+	Server     *btcec.PublicKey
+	ClaimLeaf  Leaf
+	RefundLeaf Leaf
```
altafan: replace with "raw data":
- claim key (if nil, the keyRef passed in in NewContract is used, ie we're the claimer)
- refund key (if nil, the keyRef passed in NewContract is used, ie we are the funder)
- refund locktime (absolute locktime, must always be present regardless the direction of the swap)

contract/handlers/vhtlc/handler.go
```diff
@@ -0,0 +1,545 @@
...
+		return "", fmt.Errorf("failed to get server info: %w", err)
+	}
+	return info.CheckpointTapscript, nil
+}
+
+func (h *Handler) getCheckpointExitPath(c types.Contract) ([]byte, error) {
+	checkpointExitPath, err := requireParam(c, paramCheckpointExitPath)
+	if err != nil {
+		return nil, err
+	}
```
louisinger: when is it empty ? it seems to checkpointExitPath is always set, do we really need that wrapper ?

contract/handlers/htlc/handler.go
louisinger: +1
I do think we need a `htlc.NewHTLCScriptFromOpts` like we have the `vhtlc.NewVHTLCScriptFromOpts` in vhtlc's handler

contract/handlers/vhtlc/handler.go
```diff
@@ -0,0 +1,545 @@
...
+		return nil, fmt.Errorf("vhtlc contract %s: invalid %s hex: %w", c.Script, key, err)
+	}
+	pub, err := btcec.ParsePubKey(buf)
+	if err != nil {
+		return nil, fmt.Errorf("vhtlc contract %s: invalid %s: %w", c.Script, key, err)
+	}
+	return pub, nil
+}
+
+func parseOwnerKey(c types.Contract) (*btcec.PublicKey, error) {
```
louisinger: we always encode the key as 33 bytes (https://github.com/arkade-os/go-sdk/pull/191/changes#diff-c015a847ce450c3b322c078cbd1124ed4e419bf485e9d42457a260b99a8fc270R208). so that parser seems overkill to me

contract/handlers/vhtlc/handler.go
```diff
@@ -0,0 +1,545 @@
...
+	paramPreimageHash                    = "preimageHash"
+	paramRefundLocktime                  = "refundLocktime"
+	paramClaimDelayType                  = "claimDelayType"
+	paramClaimDelayValue                 = "claimDelayValue"
+	paramRefundDelayType                 = "refundDelayType"
+	paramRefundDelayValue                = "refundDelayValue"
+	paramRefundWithoutReceiverDelayType  = "refundWithoutReceiverDelayType"
+	paramRefundWithoutReceiverDelayValue = "refundWithoutReceiverDelayValue"
+	paramNICReceiverPkScript             = "nicReceiverPkScript"
+	paramNICIntrospectorPubKey           = "nicIntrospectorPubKey"
```
louisinger: let's rename all "introspector" to "emulator"

contract/handlers/vhtlc/handler.go
```diff
@@ -0,0 +1,545 @@
...
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/txscript"
+)
+
+// Param keys stored in Contract.Params.
+const (
+	paramOwnerKeyID                      = "ownerKeyId"
+	paramOwnerKey                        = "ownerKey"
+	paramSender                          = "sender"
+	paramReceiver                        = "receiver"
```
altafan: this way, we end up duplicating one of the receiver/sender pubkeys in db. We can avoid this by storing these params instead:

```suggestion
	paramSenderKeyID                      = "senderKeyId"
	paramSender                          = "sender"
	paramReceiverKeyID                      = "receiverKeyId"
	paramReceiver                        = "receiver"
```

the idea is that only one betweeen sender and receiver key id can be set and this way we can understand what's our role in this vhtlc

contract/handlers/vhtlc/handler.go
```diff
@@ -0,0 +1,545 @@
...
+	paramServer                          = "server"
+	paramPreimageHash                    = "preimageHash"
+	paramRefundLocktime                  = "refundLocktime"
+	paramClaimDelayType                  = "claimDelayType"
```
altafan: no need for this kind of params at all. We have a util function in go-sdk/client lib to parse a locktime from int64 (btw: if val < 512 it's block type, otherwise it's seconds). Drop all these param*DelayType, the constants for tags and parser function(s))

test/e2e/swap_test.go
```diff
@@ -0,0 +1,1553 @@
...
+
+// =============================================================================
+// Refund Path Tests (Mock Boltz)
+// =============================================================================
+
+// TestChainSwapMockArkToBTCCooperativeRefund tests the cooperative refund path.
+// After the user sends the VTXO, we inject a swap.expired event to trigger
+// the refund flow. With the default refundMode=success, the cooperative refund
+// via mock-boltz's /refund/ark endpoint should be attempted.
+func TestChainSwapMockArkToBTCCooperativeRefund(t *testing.T) {
```
louisinger: it seems to me we also miss the BTC refund case (musig2 tapkey path)

contract/handlers/htlc/handler.go
```diff
@@ -0,0 +1,332 @@
...
+			return nil, fmt.Errorf(
+				"htlc contract %s: expected exactly one of %q or %q",
+				c.Script, paramClaimKeyID, paramRefundKeyID,
+			)
+		}
+		return nil, fmt.Errorf(
+			"htlc contract %s: missing wallet key ID: expected %q or %q",
+			c.Script, paramClaimKeyID, paramRefundKeyID,
+		)
+	}
```
altafan: this check is absolutely correct, but should be perfomed ONLY when we create the contract, here in GetKeyRef it's just out of scope

sekulicd: purpose of this was to protect against passed corrupted contract, reverting it ...

PR #193, opened 2026-06-05
Limit settlement batches to 50 coins

sekulicd:
## Summary

- Paginate `Settle` into multiple settlement batches when total VTXO + boarding UTXO inputs exceed 50 coins
- Log the total number of settlement batches and non-final commitment txids during multi-batch settlement
- Add internal tests for mixed VTXO/UTXO, VTXO-heavy, and UTXO-only batching

@altafan please review.

## Summary by CodeRabbit

* **New Features**
  * Settlement operations now automatically batch large input counts for improved transaction handling and efficiency.
  * Added batch-level progress logging and enhanced error reporting to clearly identify issues during multi-batch settlement operations.


ghost (changes requested):
## 🔴 Arkana Code Review — Request Changes

### Critical: Silent fund truncation is a money-loss vector

**`batch_session.go:20-22`** — The core change silently drops VTXOs beyond 50:

```go
if len(vtxos) > maxVtxosPerBatch {
    vtxos = vtxos[:maxVtxosPerBatch]
}
```

This is **protocol-critical** and has several problems:

#### 1. Silent partial settlement — users lose funds
`Settle()` returns success after settling only 50 of N VTXOs. The caller has no indication that funds were left behind. If the user believes all funds are settled (the only reasonable interpretation of a successful `Settle()` call), they may never call `Settle()` again. The remaining VTXOs will eventually expire and require costly unilateral exits — or be lost entirely if the user doesn't monitor expiration.

At minimum, the function should either:
- Loop internally until all VTXOs are settled (the batching approach from commit `672b076` that was reverted)
- Return metadata indicating how many VTXOs remain unsettled
- Return an error/warning when truncation occurs

#### 2. No expiry-aware ordering before truncation
`getSpendableVtxos()` returns VTXOs from the store without sorting by expiry. The truncation takes the first 50 arbitrarily. VTXOs closest to expiration could be at index 51+ and get dropped, which is the worst possible outcome — those are the ones that *most* urgently need settlement.

If truncation stays, VTXOs **must** be sorted by ascending expiry time before slicing.

#### 3. Boarding UTXOs are uncapped
The constant is `maxVtxosPerBatch` and only caps VTXOs, but UTXOs (boarding inputs) are passed through uncapped. If the input-count limit is a server-side constraint, UTXOs count toward it too. If someone has 60 boarding UTXOs and 30 VTXOs, this cap doesn't help.

#### 4. `CollaborativeExit` is not capped
`CollaborativeExit` (line 70+) also fetches all VTXOs via `getSpendableVtxos` and passes them to the server. If the 50-input limit is a server constraint, `CollaborativeExit` will hit the same failure this PR is trying to prevent.

### Minor

**`test/e2e/batch_session_test.go:311`** — `leftover` constant is declared but never used. The test verifies balance is still 60k and that the earliest expiry hasn't changed, but doesn't assert that exactly 10 VTXOs remain unsettled (e.g., by checking `len(bobBalance.OffchainBalance.Details)`). This would make the test more robust.

**`test/e2e/batch_session_test.go:332-339`** — The `wg.Go()` + `SendOffChain` pattern runs `NotifyIncomingFunds` concurrently with `SendOffChain` but doesn't synchronize their start order. This could be flaky if `SendOffChain` completes before `NotifyIncomingFunds` registers.

### Recommendation

The reverted batching approach (commit `672b076`) was architecturally correct — it looped over batches so the user's single `Settle()` call actually settled everything. The revert to silent truncation trades correctness for simplicity. I'd recommend either:

1. **Restore the batching loop** (preferred — settles all funds in one call)
2. **Sort by expiry + return unsettled count** (acceptable — at least prioritizes urgent VTXOs and informs the caller)

⚠️ **Protocol-critical**: This touches VTXO settlement. Requesting human review regardless of outcome.

---
*🤖 Arkana — automated review*

PR #196, opened 2026-06-12
Signer Key rotation 

sekulicd:
## Why

The arkd introduced signer key rotation. Every VTXO/boarding output embeds the server signer key in its
tapscript, so after rotation:

- **Pre-rotation outputs become invisible to restore.** Discovery derived candidate scripts
  only with the *current* signer; a wallet restored from seed after rotation could not see
  funds locked to the old key — they appeared lost.
- **Old-signer outputs are on a deadline.** The server co-signs spends of deprecated-signer
  outputs only until each key's `cutoffDate` (and only if every output of the tx commits to
  the current signer — arkd#822). After cutoff they are unilateral-exit-only.

This PR makes go-sdk rotation-aware end to end: pre-rotation funds are **discoverable,
spendable, and automatically migrated** to the current signer before cutoff.

### 1. Multi-signer discovery (`contract/`)
- `acceptedSigners(info)`: current + deprecated signer set from `GetInfo`
  (`DeprecatedSignerPubKeys`), x-only normalized, deduped; malformed entries skipped with a
  warning — a bad server entry can never fail Unlock.
- `CandidateContracts` added to the `Handler` interface (+ `DefaultCandidateContracts`
  fallback so custom handlers compile unchanged): derives one contract per accepted signer
  per owner key, offchain **and** boarding.
- `scanContracts` reworked: candidates for every signer, deduped by scriptHex, probed in a
  **single batched** indexer call per index-batch (scan cost does not grow with signer count);
  used contracts persisted with the **matching** signer in their params — so all downstream
  signing/reconstruction (which reads stored params) works with zero further special-casing.
- **Union gap counter**: the scan terminates only after `gapLimit` consecutive indices with
  no hit for *any* accepted signer. Sound because allocation is one monotonic HD index stream
  regardless of signer. (Per-signer counters — the original design — lose post-rotation funds
  on restore: with contracts at 0..1000 under the old key and 1001+ under the new key, the
  new key's scan would terminate around index 20. Regression test
  `TestScanContractsHighCurrentSignerStartIdx` fails on the old logic.)
- Single-key path: deprecated candidates probed too; allocation now filters stored contracts
  by the **current** signer (prevents reusing a discovered deprecated-signer contract for new
  addresses).
- `infoCache.Invalidate()`: discovery always starts from a fresh signer set at restore.
- `AddContract` is idempotent (`INSERT OR IGNORE` on the `script` PK) — re-scans are safe;
  `(index, old_signer)` and `(index, new_signer)` coexist as distinct rows. **The HD index
  never resets on rotation.**

### 2. Migration (`rotation.go`)
- `reconcileDeprecatedSigners`: classifies each signer with contracts as
  `Active` / `ToMigrate` (deprecated, pre-cutoff; `cutoffDate==0` = no cutoff) / `Expired`
  (past cutoff → surfaced exit-only, never sent to a collaborative settle); unknown signers
  are logged, not classified.
- Migrates **only** the to-migrate vtxos via the new `WithSettleVtxos(...)` settle option
  (subset settle — uses client-lib's `WithFunds` bypass); the change output is built from
  the current signer, so every output of the migration tx commits to the new key (#822).
- Idempotent: migrated vtxos are spent; re-runs are no-ops. Empty-subset guard prevents an
  accidental full-settle fallback.
- `DeprecatedSignerStatus` (counts per state, `NearestCutoff`, `AmountAtRisk`) exposed via
  `DeprecatedSignerSummary()`.

### 3. Sync-gated unlock (wallet usable only after migration)
- Migration runs **synchronously in the Unlock flow** after `ScanContracts`/`refreshDb`,
  *before* the sync notification fires: every public operation returns `ErrIsSyncing` until
  the migration attempt completes, so a user cannot act on a half-migrated wallet.
  Internally it uses an unexported `settle()` (public `Settle` keeps its guard; exactly two
  call sites, grep-verified).
- **Never-hostage**: migration failure logs + surfaces in status and still completes sync;
  the rotation digest is then left unseeded so the periodic refresh retries automatically.
  No-rotation wallets see zero added unlock latency.
- Live rotation (long-running daemons): a signer-set digest is checked on periodic refresh;
  on change → invalidate cache → rescan → reconcile, without restart. The digest only
  advances after a successful rescan, so transient failures retry.
  Note: requires `WithRefreshDbInterval` (defaults to 0 = off).

## Testing

- **Unit** (all `-race`): multi-signer discovery, union-gap restore scenario (differential:
  fails pre-fix), per-signer persist, dedup + single-probe call count, idempotent re-scan,
  single-key discovery + post-rotation allocation, classification incl. cutoff edge cases,
  subset-settle wiring, sync-gating (public op blocked during in-flight migration),
  never-hostage + retry, rotation-detection retry.
- **Live manual test** against arkd #1097 on regtest (full report + evidence in session
  artifacts): fund under key A → rotate to key B (A deprecated) → restore from seed with a
  fresh DB → key-A vtxos rediscovered → migrated to key-B outputs (`RegisterIntent`
  accepted — intent-proof compatibility with #1097 confirmed) → past-cutoff key classified
  Expired with no settle attempted.
- **E2E** `TestRestoreAfterSignerRotation` (`test/e2e/restore_smoke_test.go`): complete and
  compiling, `t.Skip`-guarded until arkd#1097 merges (unskipping requires implementing
  `rotateServerSigner` in the test env).
- Lint (`golangci-lint`), gosec, vet, build: clean.

## Breaking changes

- `Handler` interface: new `CandidateContracts` method (use `DefaultCandidateContracts` as a
  one-line fallback for non-rotation-aware handlers).
- `Manager` interface: new `InvalidateInfoCache()` method.
- `DeprecatedSignerStatus` shape: 3-state (`Active`/`ToMigrate`/`Expired`).
- Unlock latency: wallets holding deprecated-signer vtxos wait one settle round before the
  sync notification (intended: usable only after migration).

## Out of scope / follow-ups

- VHTLC/Boltz-specific reconstruction (ts-sdk#554 covers that ecosystem side).
- Server-side validation (arkd#1097).
- Client gRPC `X-Build-Version` header: done properly in client-lib (separate PR), since
  dial-time interceptors cannot be injected from go-sdk.
- Single-signer scan warm-up optimization (restore of the `startIdx` fast path for the
  no-rotation case) and `DeprecatedSignerSummary` on the `Wallet` interface — tracked.

## Dependencies

- Requires arkd with deprecated-signer support (arkd#1097 / #822) at runtime for rotation
  behavior; pinned client-lib already decodes `deprecated_signers` (field 17) — no bump
  needed.
- Conceptual cross-check: ts-sdk#554.

@altafan @Kukks please review 

## Summary by CodeRabbit

* **New Features**
  * Added support for signer key rotation with automatic detection of server-side signer changes.
  * Enabled wallet migration of funds from deprecated signers to the current signer.
  * Introduced `DeprecatedSignerSummary` to track and report on migratable, expired, and active funds.
  * Added settlement-subset functionality via `WithSettleVtxos` option for selective fund migration.

* **Improvements**
  * Contract discovery now discovers contracts across multiple signers (current and deprecated).
  * Wallet restore/unlock now includes automatic migration during initialization.

* **Tests**
  * Added comprehensive tests for rotation detection, migration logic, and multi-signer contract scanning.


altafan:
Replaced by #199

PR #200, opened 2026-06-17
forward and export `client.SendOption` alias

louisinger:
cherry pick from https://github.com/arkade-os/go-sdk/pull/182

@altafan @sekulicd please review

## Summary by CodeRabbit

* **New Features**
  * `SendOffChain` method now accepts optional configuration parameters for enhanced flexibility when sending off-chain transactions.
  * Added support for specifying extra packets in off-chain sends through new configuration options.

### DIFF
--- a/sdk.go
+++ b/sdk.go
@@ -66,7 +66,9 @@ type Wallet interface {
 	BurnAsset(
 		ctx context.Context, assetID string, amount uint64,
 	) (string, error)
-	SendOffChain(ctx context.Context, receivers []clienttypes.Receiver) (string, error)
+	SendOffChain(
+		ctx context.Context, receivers []clienttypes.Receiver, opts ...SendOffChainOption,
+	) (string, error)
 	RegisterIntent(
 		ctx context.Context,
 		vtxos []clienttypes.Vtxo, boardingUtxos []clienttypes.Utxo, notes []string,
--- a/send.go
+++ b/send.go
@@ -11,7 +11,7 @@ import (
 )
 
 func (w *wallet) SendOffChain(
-	ctx context.Context, receivers []clienttypes.Receiver,
+	ctx context.Context, receivers []clienttypes.Receiver, extraOpts ...SendOffChainOption,
 ) (string, error) {
 	if err := w.safeCheck(); err != nil {
 		return "", err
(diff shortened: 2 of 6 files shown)


sekulicd (approved):

ghost (approved):
## Arkana Code Review — arkade-os/go-sdk#200

**Verdict: APPROVE** ✅

### Summary

This PR adds variadic `SendOffChainOption` support to `Wallet.SendOffChain()`, forwarding the SDK-level options through to `client.SendOffChain`. It also re-exports `client.WithExtraPacket` so SDK consumers don't need to import `client-lib` directly. Cherry-picked from #182.

### Analysis

**API compatibility — no breakage.** Adding a variadic `...SendOffChainOption` parameter to `SendOffChain` is backward-compatible in Go. All 49+ downstream callers across `emulator`, `bancod`, `solver`, `ark-faucet-review`, `asset-demos`, `demos`, and `introspector-review` call `SendOffChain(ctx, receivers)` without the variadic arg, and will compile without changes. No external implementations of the `Wallet` interface were found, so the interface change is safe.

**Type alias is correct.** `send_opts.go:9` — `type SendOffChainOption = client.SendOption` uses a type alias (`=`), not a new type definition. This means `client.SendOption` values are directly assignable without conversion. Correct choice for a re-export.

**Option ordering is safe.** `send.go:73` — `extraOpts` are appended *after* the internal `WithVtxos`, `WithKeys`, and `WithReceiver` options. This is the right ordering: caller-supplied options (like `WithExtraPacket`) are additive and won't clobber the internally-computed vtxos/keys/change. If the upstream `SendOption` pattern used last-write-wins semantics for some option, a malicious caller could theoretically override vtxos or keys, but that's a trust boundary the SDK already grants to its callers.

**Docker changes are infrastructure-only.** Bumping `nbxplorer:2.5.30` → `2.6.7` and `golang:1.26.3` → `1.26.4` are routine version bumps for the test environment.

### Minor observations (non-blocking)

1. **No unit test for the new option path.** The `WithExtraPacket` forwarding is a thin wrapper so this is low-risk, but an e2e test exercising `SendOffChain(ctx, receivers, WithExtraPacket(...))` would be nice for coverage. Not blocking since the upstream `client.WithExtraPacket` is presumably tested in `arkd`.

2. **Only `WithExtraPacket` is re-exported.** If `client-lib` exposes other `SendOption` constructors, callers who need them still have to import `client-lib` directly. This is fine as a deliberate surface-area choice — just noting it.

No protocol-critical concerns. No security issues. No race conditions introduced.

🤖 Reviewed by Arkana

PR #207, opened 2026-07-09
init.go: set dust on Unlock

### LINKED ISSUE #206: wallet.SendOffChain drops asset dust carrier on reloaded wallet (dustAmount not restored on Unlock)
### Summary

`dustAmount` is only ever set in `Init()` (`init.go:79`, `w.dustAmount = info.Dust`). On the `LoadWallet` → `Unlock` path it is never re-fetched, so it stays `0`. Any code relying on `w.dustAmount` after a restart silently misbehaves.

Most visible symptom: asset sends fail after a wallet restart with `output amount is lower than min utxo amount`.

### Root cause

`SendOffChain` auto-fills the BTC dust carrier for asset receivers (`send.go:36-43`):

```go
for i, receiver := range clone {
    if len(receiver.Assets) > 0 && receiver.Amount < w.dustAmount {
        clone[i].Amount = w.dustAmount
    }
}

On a reloaded wallet w.dustAmount == 0, so the guard is receiver.Amount < 0 → always false → no carrier attached → the server rejects the sub-dust asset output.

Unlock (init.go:85-) restores the contract manager, broadcasters, sync state, etc. but never calls GetInfo, so dustAmount (and anything else derived from server info) is left at its zero value.

Reproduce

1. NewWallet + Init + Unlock, send an asset → works (dust filled).
2. Restart: LoadWallet + Unlock (no Init), send the same asset → fails with output amount is lower than min utxo amount.

Impact

Also affects other dustAmount consumers on the load path — rotation.go:178,244 (buildReceiverForMigration, sub-dust recoverable detection at rotation.go:529), which silently change behavior when dustAmount == 0.

Suggested fix

Fetch and set dustAmount in Unlock (or lazily on first use) the same way Init does:

info, err := w.client.GetInfo(ctx)
if err != nil { return err }
w.dustAmount = info.Dust

Consider consolidating whatever else Init derives from GetInfo so Load+Unlock and Init end up in the same state.

### PR DESCRIPTION
**Pull request by louisinger**
it closes #206 

@sekulicd @altafan please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet unlock reliability by updating wallet details immediately after a successful unlock, ensuring settings stay in sync with the latest configuration.
  * Refreshed wallet state values (including dust threshold and network/signer set metadata) from the server-provided data to reduce balance and signer-related inconsistencies.

### DIFF
--- a/init.go
+++ b/init.go
@@ -96,6 +96,12 @@ func (w *wallet) Unlock(ctx context.Context, password string) error {
 		return err
 	}
 
+	info, err := w.Client().GetInfo(ctx)
+	if err != nil {
+		return err
+	}
+	w.dustAmount = info.Dust
+
 	w.logMu.Lock()
 	log.SetLevel(log.DebugLevel)
 	if !w.verbose {
--- a/init.go
+++ b/init.go
@@ -2,6 +2,7 @@ package arksdk
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"sort"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by sekulicd — APPROVED** [review]

File: init.go [hunk 1]
```diff
@@ -96,6 +96,12 @@ func (w *wallet) Unlock(ctx context.Context, password string) error {
 		return err
 	}
 
+	info, err := w.Client().GetInfo(ctx)
+	if err != nil {
+		return err
+	}
+	w.dustAmount = info.Dust
```
altafan: NO! The dust amount can and should be fetched from config data

PR #214, opened 2026-07-24
Tidy up management of vhtlc contracts & Refactor SwapManager & Other fixes

altafan:
This refactors the management of the vhtlcs: before this, vhtlc contracts were imported in the contract manager only via the `SwapHandler` (now `SwapManager`) and all the logic to spend vtxos locked by such contracts lived in this very component. Now, instead, new APIs have been added to the `Wallet` interface to manage vhtlc: create/delete/list/claim/refund. We may want to move them to the contract manager in the future but for now it's just the right place as the `Wallet` **INTENTIONALLY** doesn't keep track of vtxos locked by vhtlc contracts (left TODOs). This will come along with the contract watcher component.

The `SwapManager`, since completely not used by anybody, went through a huge refactor but the original feature set has been maintained and extended: from an external POV, the notable changes are the renaming of the original `GetInvoice` to `ReverseSwap` ,  and `PayInvoice | PayOffer` merged to `SubmarineSwap`.
Internally, the APIs now consume the Wallet's new vhtlc APIs to create/solve swaps and they take care of CRUD operations to persist swap state and to trigger recover/restore of pending swaps.

BONUS:
- This renames `Settlement` to `Refresh` almost everywhere in the codebase and fixes some bugs that have been inherited when imporitng the feature from fulmine and caused degradation of the arkade-regtest env.
- Added a periodic check on schedule the next refresh as fallback strategy (also here imported from fulmine and fixed).
- The `WithoutAutoSettle` option has been dropped in favor of a `WithScheduleRefreshInterval` that can be set to big values in tests instead of disabling the feature

## Summary by CodeRabbit

* **New Features**
  * Added wallet support for creating, listing, claiming, refunding, and deleting VHTLCs.
  * Added swap management for submarine, reverse, and chain swaps, including restart and restore recovery.
  * Added identity xpub retrieval.
  * Added contract enable/disable operations with transactional bulk updates.
  * Added renewal scheduling by time or block height.

* **Improvements**
  * Replaced auto-settlement terminology/behavior with automatic renewal.
  * Added “next renewal” timing query for scheduling visibility.

* **Documentation**
  * Updated wallet and scheduler docs for auto-renewal and removed the auto-settle option references.


tiero:
I disagree on calling refresh, AI agents may then pick it up or get confused. The [term](https://docs.arkadeos.com/wallets/advanced/vtxo-management#vtxo-renewal) we use everywhere is Renew or Renewal (if the goal having a high-level explicit action)

altafan:
> I disagree on calling refresh, AI agents may then pick it up or get confused. The [term](https://docs.arkadeos.com/wallets/advanced/vtxo-management#vtxo-renewal) we use everywhere is Renew or Renewal (if the goal having a high-level explicit action)

yeah i had the doubt refresh/renew, i flipped the coin and got unlucky. Fixing.

PS. worst motivation ever 😂

altafan:
@tiero Renaming done in [e7de0b4](https://github.com/arkade-os/go-sdk/pull/214/commits/e7de0b4909d0d5d1e48912484cd25593d0cbe294)

tiero:
utACK

store/sql/contract_store.go
```diff
@@ -160,6 +168,37 @@ func (v *contractStore) Clean(ctx context.Context) error {
...
+	}
+
+	update := func(querier *queries.Queries) error {
+		for _, script := range scripts {
+			n, err := querier.UpdateContractState(ctx, queries.UpdateContractStateParams{
+				State:  string(state),
+				Script: script,
+			})
+			if err != nil {
+				return err
```
bitcoin-coder-bob: if we hit this error the caller will not get back info relating to which script we errored on. Can we include that in the error like we do on line 190

identity/utils.go
```diff
@@ -10,12 +10,13 @@ import (
...
 	"strings"
 
+	arklib "github.com/arkade-os/arkd/pkg/ark-lib"
 	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/chaincfg"
 	"golang.org/x/crypto/pbkdf2"
 )
 
-func getBIP86RootPath(network chaincfg.Params) []uint32 {
+func getBIP86RootPath(network *chaincfg.Params) []uint32 {
```
bitcoin-coder-bob, line 19: do we need a nil check now? If so then we should probably return an error type as well

bitcoin-coder-bob: I know we call this function with the result from `toBitcoinNetwork` which should not yield a nil value, but this will make us future proof safe if it is called elsewhere with a nil value or if `toBitcoinNetwork` is ever changed to return a nil value

identity/identity.go
bitcoin-coder-bob, line 266: I know you didnt change this but caught my eye, if we pass an `id` value of less than 0 should we error?

init.go
```diff
@@ -254,46 +259,98 @@ func (w *wallet) IsLocked(_ context.Context) bool {
...
+	}
+	log.Debugf("scheduled next renewal at %s", nextExpiration.Format(time.RFC3339))
+}
+
+// periodicScheduleNextRenewal recomputes the next renewal from the full vtxo set at a
+// fixed interval, as a safety net: it recovers the auto-renewal loop if a vtxo event was missed
+// or a previous renewal couldn't be scheduled (eg. an already-expired vtxo appeared while no
+// event was fired).
+func (w *wallet) periodicScheduleNextRenewal(ctx context.Context) {
+	if w.scheduler == nil || w.renewVtxosScheduleInterval == 0 {
```
bitcoin-coder-bob, line 359: lets add a check to ensure `renewVtxosScheduleInterval` is of some minimal value or at least >=0

init.go
```diff
@@ -254,46 +259,98 @@ func (w *wallet) IsLocked(_ context.Context) bool {
...
+	// renewal in the past.
+	if !earliest.After(time.Now()) {
+		renewVtxos()
+		return
+	}
+
+	// Schedule the renewal slightly before the earliest expiration (proportional lead), but
+	// only if it's sooner than the one already scheduled.
+	expiry := time.Until(earliest)
+	nextExpiration := time.Now().Add(expiry * 9 / 10)
```
bitcoin-coder-bob: what do you think if we add a floor value here? if the value ends up being under the floor, we set to a floor value

bitcoin-coder-bob: potential for overflow error here. example would be if expiry is set to 280320 hours which would trigger an overflow. AI had a good suggestion reqrtie for this section of code:

```
  const minRenewalLead = 2 * time.Minute // or whatever a batch round costs

  expiry := time.Until(earliest)
  if expiry <= minRenewalLead {
      renewVtxos()
      return
  }
  lead := expiry / 10 * 9
  if expiry-lead < minRenewalLead {
      lead = expiry - minRenewalLead
  }
  nextExpiration := time.Now().Add(lead)
```
That collapses the "already expired" and "about to expire" cases into one branch, removes the overflow, and guarantees a lead that is actually long enough to complete a settle.

swap/types/types.go
```diff
@@ -0,0 +1,60 @@
...
+
+type Swap struct {
+	Id          string
+	From        boltz.Currency
+	To          boltz.Currency
+	CreatedAt   time.Time
+	UpdatedAt   time.Time
+	Status      int
+	VHTLCScript string
+	Amount      uint64
```
bitcoin-coder-bob: sats? msats? worth a comment. or amount could depend on currency type specified

swap/store/sql/utils.go
```diff
@@ -29,9 +48,9 @@ func int64Value(value sql.NullInt64) int64 {
 	return value.Int64
 }
 
-func requireAffected(count int64, format, id string) error {
-	if count == 0 {
-		return fmt.Errorf("%w: "+format, swapdomain.ErrNotFound, id)
+func decodeNullableHex(value sql.NullString) ([]byte, error) {
+	if !value.Valid {
+		return nil, nil
```
bitcoin-coder-bob, line 53: throw an error here no?

swap/scheduler/gocron/service.go
```diff
@@ -0,0 +1,213 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	s.blockCancel = cancel
+
+	go func() {
+		t := time.NewTicker(s.explorerPollInterval)
+		defer t.Stop()
+		for {
+			h, err := s.explorer.GetBlockHeight()
+			if err == nil {
```
bitcoin-coder-bob, line 59: should we do a debug log if there was an error yielded by GetBlockHeight? Otherwise I think we keep hitting iterations of the for loop without clear indication when `ctx.Done` is not hitting

swap/swap.go
```diff
@@ -5,873 +5,271 @@ import (
...
-		[]clientTypes.Receiver{{To: session.destinationAddr, Amount: session.totalAmount}},
-		withoutReceiver,
-		map[string]*vhtlc.VHTLCScript{session.vtxos[0].Script: session.vhtlcScript},
-		h.config,
-		session.signerSession,
-	)
-	if err != nil {
-		return "", fmt.Errorf("failed to setup refund batch session handler: %w", err)
-	}
+func (h *SwapManager) SubmarineSwap(
```
bitcoin-coder-bob, line 162: is this swap in a particular direction or in either direction? Worth a comment

swap/preimage.go
```diff
@@ -6,65 +6,44 @@ import (
...
-	signer sdkidentity.KeyedPreimageSigner, keyRef arkidentity.KeyRef, index uint32,
-) ([]byte, error) {
+// DerivePreimage generates a deterministic preimage for Boltz swaps by crafting and signing a
+// message with a well defined format (check BuildPreimageMessage) and signing its SHA256 hash.
+// The SHA256 hash of the signature is returned as the preimage.
+//
+//	message_hash	= SHA256(message)
+//	sig				= BIP340_SchnorrSign(local_key, message, aux_rand=0)
+//	preimage 		= SHA256(sig)
+func DerivePreimage(
```
bitcoin-coder-bob, line 29: I think we may run into a preimage uniqueness issue here. `BuildPreimageMessage(keyRef.PubKey, 0)` always uses `0` for now. I'm actually not clear on the blast radius of this but this smelled to me so I'm flagging it. 

I dont think we have test coverage that proves two swaps have different preimages

swap/btc_util.go
```diff
@@ -58,23 +56,23 @@ func constructClaimTransaction(
...
 
 	vbytes := computeVSize(tx)
 
-	feeRate, err := explorerClient.GetFeeRate()
+	feeRate, err := explorerSvc.GetFeeRate()
 	if err != nil {
 		return nil, err
 	}
 
 	feeAmount := uint64(math.Ceil(float64(vbytes)*feeRate) + 100)
```
bitcoin-coder-bob, line 75: `100` feels like a magic number, lets add a comment

bitcoin-coder-bob (changes requested):

vhtlc.go
```diff
@@ -0,0 +1,892 @@
...
+	vtxo, pending, err := w.selectClaimableVTXO(ctx, vhtlcContract.Script, o.outpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	args, err := handler.GetArgs(vhtlcContract)
+	if err != nil {
+		return nil, err
+	}
+	parsed, ok := args.(vhtlchandler.ContractArgs)
```
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: Blocker**

`GetArgs` on a non-interactive contract returns `NonInteractiveContractArgs` (handlers.go:170), not `ContractArgs`. Embedding does not make this type assertion succeed, so for every contract created via `CreateVHTLC` with `NonInteractiveEmulator`/`NonInteractiveReceiver` set, all three of `ClaimVHTLC`, `RefundVHTLC` and `UnilateralRefundVHTLC` fail with `invalid contract args type`.

`vhtlcScript()` at line 673 already handles this correctly by branching on `contract.Type`. This function needs the same branch.

No test catches it: `TestNonInteractiveClaim` builds the script by hand and never goes through `CreateVHTLC` or the wallet claim path.

swap/swap.go
```diff
@@ -1034,59 +419,50 @@ func (h *SwapHandler) submarineSwap(
...
 				return nil, err
 			}
 			go func() {
-				if err := unilateralRefund(*swapDetails); err != nil {
-					log.WithError(err).Errorf("failed to refund swap %s unilaterally", swap.Id)
+				log.Debugf(
+					"process aborted while waiting for updates for swap %s, "+
+						"trying to schedule a unilrateral refund before aborting...", swapResp.Id,
+				)
+				if err := h.scheduleUnilateralRefund(ctx, swap, parsed.RefundLocktime); err != nil {
```
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: Blocker**

`ctx` here is the timeout context created at line 401, and this branch only runs when that context is already done. `scheduleUnilateralRefund` captures it in the `refund` closure and calls `UnilateralRefundVHTLC(ctx, ...)` when the task eventually fires, so the refund always fails with `context canceled`.

`chainswap.go:308` has the identical pattern. `RefundSubmarineSwap` is a milder version: it hands a request-scoped context to a task that may not fire until the refund locktime, potentially hours later.

These closures should capture `h.ctx`, the manager lifetime context, not the request or timeout one.

swap/swap.go
```diff
@@ -1164,153 +536,105 @@ func (h *SwapHandler) reverseSwap(
...
+		CreatedAt:   time.Now(),
+		Status:      int(SwapStatusPending),
+		VHTLCScript: contract.Script,
+		Amount:      swapResp.OnchainAmount,
+		LNSwap: &swaptypes.LNSwapInfo{
+			Invoice:      swapResp.Invoice,
+			PreimageHash: preimageHashHASH160,
+		},
+	}
+	if persistPreimage {
```
bitcoin-coder-bob, line 609: Bob had the AI generate this comment.

**Severity: Blocker**

The preimage is persisted exactly when it is also re-derivable, and dropped when it is random and therefore unrecoverable.

`persistPreimage` is the `PreimageSigner` type assertion, so:
- identity supports schnorr: preimage is deterministic, `swapPreimage` in recover.go can re-derive it, and it is *also* written to the store (redundant);
- identity does not: preimage comes from `rand.Read`, cannot be re-derived, and is *not* written.

For the second case a crash between the invoice being paid and the claim landing loses the funds outright. We are the receiver, so there is no refund path. Same pattern at `chainswap.go:184`.

This one needs no attacker and no unusual user action to trigger, which is why I would rank it above the key-reuse issue.

vhtlc.go
```diff
@@ -0,0 +1,892 @@
...
+	}
+
+	return arkTxid, nil
+}
+
+func (w *wallet) DeleteVHTLCs(ctx context.Context, scripts []string) error {
+	if err := w.safeCheck(); err != nil {
+		return err
+	}
+	return w.store.ContractStore().DisableContracts(ctx, scripts)
```
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: Blocker**

Two problems on this line.

**1. It rewinds the VHTLC key counter, which reuses preimages.** `SelectLatestActiveContractByType` is `WHERE state = 'active' ORDER BY key_index DESC LIMIT 1`. Disabling the highest-key-index active VHTLC contract makes the next `getNewKey` return that same index, so the same key, so the same derived preimage, so the same payment hash.

Blast radius, for the two flows where we choose the preimage (submarine swaps are unaffected, their hash comes from the invoice):
- **Ark to BTC chain swap.** Boltz is the vhtlc receiver, so the claim leaf is their signature plus the preimage. If they already know the preimage from the earlier swap at that key, they claim your vhtlc the moment you fund it and never lock the BTC. Direct loss.
- **Reverse swap.** Boltz can settle the incoming LN HTLC without ever funding your vhtlc. Loss of the swapped amount.

The `contractMu` comment says a reused key means "a duplicate payment hash rejected by Boltz". Rejection is the lucky outcome, and relying on the counterparty's dedup for your own fund safety is the wrong place for that control.

Worth noting this is genuinely new here: before this PR the only path setting a contract inactive was signer rotation over migrated (default) contracts, so the VHTLC counter was effectively monotonic.

**2. No validation that these are VHTLC contracts.** Pass a default wallet script and it goes inactive; `refreshDb` and `listenForArkTxs` then skip it and the user silently stops tracking their own vtxos. It also bypasses `contractManager.DisableContracts`, the method this PR added for exactly this, with the manager mutex held.

swap/swap.go
```diff
@@ -1342,278 +666,188 @@ func (h *SwapHandler) waitAndClaim(
...
+	return nil
 }
 
-func (h *SwapHandler) getNewKey(
+func (h *SwapManager) getNewKey(
 	ctx context.Context, handler handlers.Handler,
 ) (*identity.KeyRef, error) {
-	store := h.arkWallet.Store().ContractStore()
-	keyProvider := h.arkWallet.Identity()
+	store := h.wallet.Store().ContractStore()
```
bitcoin-coder-bob, line 895: Bob had the AI generate this comment.

**Severity: High**

This is the fix site for the key-reuse issue flagged on `vhtlc.go:331`. To be clear about provenance: this function is byte-identical to master, so it is not a regression, but `DeleteVHTLCs` makes its weakness reachable for the first time.

Deriving from "latest **active** contract" means any deactivation rewinds the counter. Suggest a monotonic high-water mark instead: `MAX(key_index)` across all contracts regardless of state, or a dedicated counter. That fixes this and `contractManager.newVHTLCContract` together, and it also closes the restore path where Boltz's gap-limit-50 scan misses old swaps and hands back a rewound counter.

Separately (pre-existing, low priority): `newVHTLCContract` keeps per-type counters but both types index into the same `m/0/i` keyspace, so VHTLC #3 and non-interactive VHTLC #3 share a pubkey. That is address linkage rather than a preimage collision, but it argues for the same fix.

init.go
```diff
@@ -254,46 +259,98 @@ func (w *wallet) IsLocked(_ context.Context) bool {
...
+	}
+	// Nothing to settle.
 	if len(vtxos) <= 0 {
 		return
 	}
 
-	sort.SliceStable(vtxos, func(i, j int) bool {
-		return vtxos[i].ExpiresAt.Before(vtxos[j].ExpiresAt)
-	})
+	renewVtxos := func() {
```
bitcoin-coder-bob, line 307: Bob had the AI generate this comment.

**Severity: High**

`renewVtxos` is invoked synchronously at line 320 when the earliest vtxo is already expired, while `scheduleMu` is held. `scheduleNextRenewal` is called from `listenForArkTxs` (wallet.go:992 and 1003), so a full batch session runs on the ark tx event listener goroutine.

Two consequences: event processing stalls for the duration of the settle, and `Stop()`'s `waitForBackground(5 * time.Second)` times out. It also uses `context.Background()`, so it is not cancelled by `Lock`/`Stop`.

The swap scheduler already has the right instinct for this, `swap/scheduler/gocron/service.go:155`: "Run async so callers holding their own locks (e.g. the vtxo event listener) don't deadlock."

init.go
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: Medium**

Three things wrong with this arithmetic.

**No floor on the lead.** The 90% rule is purely proportional. A vtxo one second from expiry schedules a renewal 900ms out, and `renewVtxos` runs a whole batch session, so it completes well after expiry. Any vtxo whose remaining life is shorter than a batch round gets a renewal guaranteed to be too late, silently.

**TOCTOU against the guard at line 317.** That check reads the clock, `time.Until` at 324 reads it again. If `earliest` falls between the two, `expiry` is <= 0, `nextExpiration` lands in the past, and `ScheduleTask` returns "cannot schedule task in the past". You get a warning and nothing scheduled for that cycle. It self-heals on the next periodic tick, so it is a skipped cycle rather than a wedge.

**Overflow.** Go parses this as `(expiry*9)/10` and `time.Duration` is int64 ns, so it wraps above ~32.5 years: `expiry=289080h` gives `lead=-252237h`. Needs a corrupt `ExpiresAt` to hit, but `expiry / 10 * 9` costs nothing.

Measured the first point empirically (20k iterations): at `expiry=1µs`, 4/20000 runs produced `nextExpiration >= earliest`; at 1ms and above, 0/20000. The break-even is `(t_C - t_B) < expiry/10`, so a 10ms GC pause breaks anything under ~100ms.

Suggested shape:

```go
const minRenewalLead = 2 * time.Minute // or whatever a batch round costs

expiry := time.Until(earliest)
if expiry <= minRenewalLead {
    renewVtxos()
    return
}
lead := expiry / 10 * 9
if expiry-lead < minRenewalLead {
    lead = expiry - minRenewalLead
}
nextExpiration := time.Now().Add(lead)
```

That collapses the expired and about-to-expire cases into one branch, removes the overflow, and guarantees a lead long enough to finish a settle.

Minor: `earliest` is `vtxo.ExpiresAt` deserialised from the server, so it carries no monotonic reading and these comparisons fall back to wall clock. An NTP step or a resume from suspend moves a renewal to the wrong side of expiry by far more than any of the above.

swap/chainswap.go
```diff
@@ -528,703 +111,400 @@ func (h *SwapHandler) ChainSwapArkToBtc(
...
+				swap.Status = int(SwapStatusFailed)
+				if err := h.persistUpdatedSwap(context.Background(), *swap); err != nil {
+					log.WithError(err).Errorf("failed to update swap %s", swapResp.Id)
+				}
+
+				log.Debugf(
+					"process aborted while waiting for updates for swap %s, "+
+						"trying to schedule a unilrateral refund before aborting...", swapResp.Id,
+				)
+				if err := h.scheduleUnilateralRefund(ctx, swap, parsed.RefundLocktime); err != nil {
```
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: High**

Same cancelled-context bug as `swap.go:450`. This branch only runs when `ctx` is done, and `scheduleUnilateralRefund` captures it for a task that fires later, so the refund always fails with `context canceled`. Should capture `h.ctx`.

swap/swap.go
```diff
@@ -924,104 +322,91 @@ func (h *SwapHandler) submarineSwap(
...
-	}
-
 	if err := contractManager.ImportContract(ctx, *contract); err != nil {
 		return nil, err
 	}
 	unlockContractMu()
 
 	ws := h.boltzSvc.NewWebsocket()
-	if err := ws.ConnectAndSubscribe(ctx, []string{swap.Id}, 5*time.Second); err != nil {
+	if err := ws.ConnectAndSubscribe(ctx, []string{swapResp.Id}, 5*time.Second); err != nil {
```
bitcoin-coder-bob, line 381: Bob had the AI generate this comment.

**Severity: Medium**

This websocket is never closed. There is no `defer ws.Close()` on any return path out of `submarineSwap`, and the same applies to `chainswap.go:163` in `ArkadeToBtcChainSwap`.

`waitAndClaim` gets it right at line 641, worth copying that.

swap/swap.go
```diff
@@ -924,104 +322,91 @@ func (h *SwapHandler) submarineSwap(
...
+		Status:      int(SwapStatusPending),
+		VHTLCScript: contract.Script,
+		Amount:      swapResp.ExpectedAmount,
+		FundingTxid: txid,
+		LNSwap: &swaptypes.LNSwapInfo{
+			Invoice:      invoice,
+			PreimageHash: preimageHash,
+		},
+	}
+	if err := h.persistSwap(ctx, *swap); err != nil {
```
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: Medium**

The swap is persisted *after* `SendOffChain` funds the vhtlc (line 368). A crash in between leaves funds locked in a vhtlc with no local record.

Recovery does not cover it: `restoreSwaps` only runs when the store is completely empty (restore.go:36), so with any other swap present that orphan is never restored.

`ArkadeToBtcChainSwap` gets the ordering right, persisting first and funding in the background. Suggest persisting as pending before funding here too.

swap/opts.go
```diff
@@ -0,0 +1,89 @@
...
+
+// ApplyOptions applies options to a new default opts and returns the first error encountered,
+// if any. Exposed for use in external (swap_test) test packages.
+func ApplyOptions(options ...Option) error {
+	_, err := applyOptions(options...)
+	return err
+}
+
+const (
+	defaultTimeout      = 15 * time.Second
```
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: Medium**

15s is too short to be usable as a default. It is used both as Boltz's `PaymentTimeout` and as the local context deadline for the entire flow, including `ArkadeToBtcChainSwap`, which needs an on-chain BTC lockup. At 15s chain swaps are essentially guaranteed to time out and route to the refund path.

The e2e tests pass `300*time.Second`, which is the tell. Previously the timeout was a required constructor argument so there was no default to get wrong. Suggest a default in the minutes, and separate defaults for LN versus chain swaps.

Also on this file: `WithTimeout`'s doc comment at line 23 is a copy of `WithOutpoint`'s and describes the wrong thing, and `defaulOpts` at line 74 is missing a `t`.

swap/restore.go
```diff
@@ -0,0 +1,283 @@
...
+		keyRef, err := keyProvider.GetKey(ctx, keyId)
+		if err != nil {
+			return nil, err
+		}
+		return &boltz.RestoreSwapsRequest{
+			PublicKey: hex.EncodeToString(keyRef.PubKey.SerializeCompressed()),
+		}, nil
+	}
+
+	xpub, err := keyProvider.GetXpub(ctx)
```
bitcoin-coder-bob, line 97: Bob had the AI generate this comment.

**Severity: Medium**

`NewSwapManager` always kicks off `restoreSwaps` in the background (swap.go:137), which sends the account xpub to Boltz. That lets them derive and link every key in the branch, including contracts that have nothing to do with swaps, and there is no way to turn it off.

Suggest a `WithoutRestore()` or opt-in `WithRestore()` option. Defaulting an SDK consumer into third-party xpub disclosure is a surprising default.

swap/store/sql/swap_store.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if swap.LNSwap != nil {
+		params.LnPreimageHash = nullableString(hex.EncodeToString(swap.LNSwap.PreimageHash))
+		params.LnInvoice = nullableString(swap.LNSwap.Invoice)
+	}
+	if swap.ChainSwap != nil {
+		params.ChainFundingTxid = nullableString(swap.ChainSwap.FundingTxid)
+		params.ChainRedeemTxid = nullableString(swap.ChainSwap.RedeemTxid)
+		params.ChainAddress = nullableString(swap.ChainSwap.Address)
+		params.ChainPrivateKey = nullableString(swap.ChainSwap.PrivateKey)
```
bitcoin-coder-bob, line 51: Bob had the AI generate this comment.

**Severity: Medium**

`chain_private_key` (the ephemeral BTC claim key) and `preimage` are written as plaintext hex into an unencrypted SQLite file. The wallet seed goes through the encrypted identity store; these do not.

Exposure is limited since the claim key is ephemeral and per-swap, but it is worth either encrypting with the same password-derived key or documenting the expectation explicitly.

swap/types/swap_handler.go
```diff
@@ -0,0 +1,64 @@
+package types
```
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: Low**

`ChainSwapEventHandler` and `SwapEventHandler` in this file (64 lines) have no implementations and no references anywhere in the tree.

Other dead code in the same package worth removing in this pass:
- `swap/store_helper.go` is now literally just `package swap`.
- `swap.ErrorNoVtxosFound` (swap.go:45) is unused, duplicates `arksdk.ErrNoVtxosFound`, and should be `ErrNoVtxosFound` per Go convention.
- `swap/scheduler/gocron/service.go:145-213`: `ScheduleNextRenewal`, `WhenNextRenewal` and `CancelNextRenewal` are not on `types.Scheduler` and `NewScheduler` returns the interface, so they are unreachable. Looks like a copy of the wallet scheduler.

Typo in this file: `OnRransactionConfirmed` at line 45.

swap/scheduler/gocron/service.go
```diff
@@ -0,0 +1,213 @@
...
+	delay := time.Until(at)
+	if delay <= 0 {
+		go task()
+		return nil
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	_, err := s.scheduler.Every(delay).WaitForSchedule().LimitRunsTo(1).Do(func() {
```
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: Low**

The job closure does `task()` and then `s.mu.Lock(); defer s.mu.Unlock()`, which locks and immediately unlocks at return. It is a no-op, presumably a leftover from something that used to clear the job.

Also in this file: the 8-field unkeyed struct literal at line 34 is fragile, and `if target <= 0` at line 99 is on a `uint32`.

test/docker/setup.sh
```diff
@@ -424,7 +424,26 @@ else
...
+boltz_ready=0
+for i in {1..30}; do
+    if curl -sf http://127.0.0.1:9001/v2/swap/submarine >/dev/null 2>&1; then
+        boltz_ready=1
+        echo "  ✅ Boltz ready"
+        break
+    fi
+    sleep 2
+done
+if [ $boltz_ready -ne 1 ]; then
```
bitcoin-coder-bob, line 449: Bob had the AI generate this comment.

**Severity: Low**

`exit` takes a numeric status, not a message, so this does not fail cleanly. Should be:

```sh
echo "  ❌ Boltz did not become ready after restart" >&2
exit 1
```

vhtlc.go
```diff
@@ -0,0 +1,892 @@
...
+		)
+	}
+
+	// Fail fast if the preimage doesn't match the contract's hash, instead of hitting an
+	// obscure server error at submit time.
+	if err := preimage.Validate(); err != nil {
+		return "", err
+	}
+	_, preimageHash := preimage.Hash()
+	if !bytes.Equal(preimageHash, spend.vhtlcOpts.PreimageHash) {
```
bitcoin-coder-bob, line 135: Bob had the AI generate this comment.

**Severity: Low**

`got` and `expected` are swapped. `preimageHash` is the computed one and `spend.vhtlcOpts.PreimageHash` is the expected one, but they are passed in the opposite order.

bitcoin-coder-bob (commented):
Bob had the AI generate this comment.

This is an AI-generated review of the branch at `e7de0b4`. I built it, ran the unit suite (all green) and ran `golangci-lint v2.9.0` (0 issues), then read through the diff. Findings are inline, each tagged with a severity.

**Summary:** the direction is good. The wallet-level VHTLC API is much cleaner than the old handler-owned logic, the `Lock`/`Stop` ordering fixes look correct, and the recover/restore work closes real gaps. There are four findings I would treat as blocking, and the scope is very large for one PR.

Two things that do not anchor to a line:

**BTC to Ark chain swaps were removed** (Severity: High, or a docs fix if intentional). `ChainSwapBtcToArk` and `RefundBtcToArkSwap` are gone, and `restoreSwap` explicitly rejects that direction at restore.go:109. The description says "the original feature set has been maintained and extended", which is not accurate. If the removal is deliberate, worth saying so in the description and changelog since it is a breaking capability loss.

**Breaking API changes that need a changelog entry** (Severity: Medium), none of which are called out in the description:
- `Wallet.WhenNextSettlement()` renamed to `WhenNextRenewal()`
- `WithoutAutoSettle()` removed
- `store.Config` moved to `types.StoreConfig`; `types.Store` gains `GetConfig()`
- `types.ContractStore` gains `DisableContracts`/`EnableContracts`, `contract.Manager` gains `DisableContracts`, so any external implementation of these interfaces breaks
- `identity.Create` now takes `arklib.Network` instead of `chaincfg.Params`
- `identity.KeyedPreimageSigner` removed
- `SwapHandler` to `SwapManager`, `GetInvoice` to `ReverseSwap`, `PayInvoice`/`PayOffer` to `SubmarineSwap`
- `swap.ClaimTransactionParams` unexported

**Test coverage gaps:** non-interactive VHTLC via `CreateVHTLC` (zero coverage, which is why the blocker on vhtlc.go:396 shipped), `ListVHTLCs` including the `WithScripts` filter, `DeleteVHTLCs` and its interaction with key derivation, and the already-expired branch of `scheduleNextRenewal`. Also `swap/opts.go` exports `ApplyOptions` purely so external tests can call it; an `export_test.go` would avoid widening the public surface.

swap/store/sql/utils.go
altafan: nono, `!Valid` means the value is null, which is totally fine since the field is optional (nullable)

swap/swap.go
altafan: I'll add comments, the direction of this swap is ARK -> LN btw

swap/preimage.go
altafan: it took me a little while to get this too, but that index is an arbitrary value, like a nonce, used to be able to derive many different preimages from the same pubkey. This is not the derivation index of the pubkey. We hardcoded it to 0 because we want 1-1 relationship between wallet pubkey and vhtlc preimage

swap/store/sql/swap_store.go
altafan: this is intended and we do not plan to change behavior soon. As the ai said, these are ephemeral sensitive info. It's like a "known limitation" that we are ok to live with, at least for now.

identity/utils.go
altafan: won't fix. This is used internally and we use the pointer instead of the struct type just because we make use of another function from btcsuite hdkeychain that requires the pointer as well.

identity/identity.go
altafan: no, the empty string id is the way to get the very first index

init.go
altafan: it can be set only via option, defaults to 30s and we already do all sanity checks there

swap/scheduler/gocron/service.go
altafan: well, out from the scope of this pr, but we should get rid of the scheduler defined in `swap/` as it's basically a bad clone of the one defined in `scheduler/` (a scheduler shouldn't know anything about renewals, it should just have primitives to implement such logic instead). Will tidy up in a dedicated pr

swap/btc_util.go
altafan: this is indeed magic number ffs, i didn't notice it before :(. Will try to fix

init.go
altafan: This could result in a client that constantly joins batches and can't be really used for other operations. For example, wit vtxo tree expiration set to 40 blocks, and therefore translated to 40s vtxo expiration, we would always be under the floor.
To prevent this, we would need to use a floor of like `10s` that would basically leave things as they are in case of delays expressed in seconds.

Won't fix.

altafan: What i'm fixing instead is the overflow by just switching operands: `/ 10 x 9` instead of `x 9 / 10`

swap/opts.go
altafan: fixed by splitting the WithTimeout option in WithLnSwapTimeout and WithChainSwapTimeout that default respectively to 2 and 30 mins

init.go
altafan: i like the renewal being done in sync, this way there can't be short period of time where WhenNextRenewal can return 0. Will fix the rest, but the floor lead.

vhtlc.go
altafan: this is fixed by dropping the active filter on select latest contract by type. Also, the key reuse shared by standard and non-interactive vhtlcs is fixed by  keeping the distinction between the 2 types only at app level. In db they are persisted with the same type, and only when reading from the db the distinction is restored. This way, when we have to derive the next key to be used for whatever vhtlc, we won't reuse any key

swap/btc_util.go
bitcoin-coder-bob: was not addressed yet

swap/swap.go
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: High (regression introduced by the fix)**

Flagging this because the `persistPreimage := ok` -> `!ok` inversion is correct for reverse swaps and for `BtcToArkadeChainSwap`, but it breaks recovery for `ArkadeToBtcChainSwap`.

That flow funds the vhtlc as the **sender**, so `ReceiverKeyId` is empty on the contract:

```go
persistPreimage := !ok        // chainswap.go:54
...
SenderKeyId:    keyRef.Id,    // chainswap.go:88  -> ReceiverKeyId is ""
```

and the re-derivation fallback bails on exactly that condition:

```go
// recover.go:338
if contractArgs.ReceiverKeyId == "" {
    return nil, fmt.Errorf("we are not the receiver of the vhtlc, cannot derive preimage")
}
```

So with the default HD identity (which does implement `PreimageSigner`, making `ok` true), the preimage for an Ark -> BTC chain swap is now **neither persisted nor re-derivable**. After a restart `recoverChainSwapClaim` (recover.go:167) fails outright and the BTC Boltz locked up can't be claimed. Before the inversion it was persisted, so this path worked.

The condition was never really "is it derivable", it is "can *this* swap re-derive it", and only the receiver-side swaps can.

### Suggested fix

Simplest correct rule is to always persist. The derivable case then just keeps a redundant copy, which is what it did before this PR and is harmless:

```go
// chainswap.go:54, chainswap.go:359, swap.go:513 -- delete the variable
- persistPreimage := !ok
  var preimage vhtlc.Preimage
  if ok { ... } else { ... }
```

```go
// chainswap.go:186, chainswap.go:478, swap.go:605
- if persistPreimage {
-     swap.Preimage = preimage
- }
+ // Always persist: the re-derivation fallback in swapPreimage only works for swaps where
+ // we are the vhtlc receiver, so sender-side ones (Ark -> BTC chain swaps) have no other
+ // way back to the preimage after a restart.
+ swap.Preimage = preimage
```

If you would rather keep the preimage out of the store when it is recoverable, the condition has to key off the role rather than the identity type, i.e. persist when we are the sender:

```go
persistPreimage := !ok || senderKeyIdIsOurs
```

but I would just persist unconditionally, it is less to reason about.

### Why CI is green

The e2e test that covers this path, `TestRecoverPendingSwaps/claims a chain swap left pending on shutdown`, uses the `createBoltzChainSwap` helper, which builds the swap record by hand and writes `Preimage: preimage` directly into the store. `swapPreimage` short-circuits on `len(swap.Preimage) > 0`, so the fallback is never exercised and the test passes against broken code.

Worth adding a test that drives `ArkadeToBtcChainSwap` itself, stops the manager mid-flight and recovers. That is the only thing that would have caught this.

vhtlc.go
bitcoin-coder-bob: Bob had the AI generate this comment.

**Severity: High**

Following up on this: the approach is good and closes both halves, but collapsing the stored type needs a data migration and `store/sql/migration/` is untouched in this PR.

For an existing wallet, rows are already persisted with `type = 'noninteractivevhtlc'`. After this change every read path queries the *storage* type `vhtlc`:

```go
// contract_store.go
rows, err := v.querier.SelectActiveContractsByType(ctx, string(storageContractType(contractType)))
...
row, err := v.querier.SelectLatestContractByType(ctx, string(storageContractType(contractType)))
```

so those pre-existing rows are:

1. never returned by `GetActiveContractsByType(ctx, ContractTypeNonInteractiveVHTLC)`, since that now looks for `vhtlc`; and
2. invisible to `GetLatestContract`, which means they do not participate in the key-index counter. That reintroduces exactly the key-reuse this change fixes, for wallets upgrading rather than starting fresh.

(Reads by script are fine: `toContract` -> `resolveContractType` leaves an old `noninteractivevhtlc` value untouched, so it round-trips.)

### Suggested fix

`store/sql/migration/20260728000000_collapse_vhtlc_contract_type.up.sql`:

```sql
UPDATE contract SET type = 'vhtlc' WHERE type = 'noninteractivevhtlc';
```

`...down.sql`, reversible because the non-interactive param is what distinguishes them:

```sql
UPDATE contract
SET type = 'noninteractivevhtlc'
WHERE type = 'vhtlc'
  AND json_extract(params, '$.nonInteractiveReceiver') IS NOT NULL;
```

If you would rather not depend on the JSON1 extension, `AND params LIKE '%"nonInteractiveReceiver"%'` works too, just less precisely.

Worth a store test that seeds a row with the old type and asserts both that `GetActiveContractsByType(NonInteractiveVHTLC)` finds it and that `GetLatestContract(VHTLC)` counts it, otherwise this regresses silently on the next schema change.

swap/chainswap.go
```diff
@@ -768,463 +416,753 @@ func (h *SwapHandler) ChainSwapBtcToArk(
...
+	if swap.From == boltz.CurrencyArk {
+		contractArgs, err := h.getSwapContractArgs(ctx, swap)
+		if err != nil {
+			return nil, nil, err
 		}
+		return h.refundSwap(swap, contractArgs)
+	}
 
-		return "", nil
+	if swap.ChainSwap.FundingTxid == "" {
```
bitcoin-coder-bob, line 687: Bob had the AI generate this comment.

**Severity: High**

A BTC lockup funded after the manager stops watching it can never be refunded, by either the automatic or the manual path.

`ChainSwap.FundingTxid` is written in exactly one place for this direction, inside the live websocket goroutine:

```go
// chainswap.go:533
swap.ChainSwap.FundingTxid = update.Transaction.Id
```

Once that goroutine exits (the `chainSwapTimeout` deadline at chainswap.go:638, or a `SwapExpired`/`TransactionFailed` at chainswap.go:615) nothing ever populates it again. And every refund path gates on it being non-empty:

- this guard in `RefundChainSwap`, the manual escape hatch;
- `recoverPendingSwaps` at recover.go:47, which skips failed swaps whose `ChainSwap.FundingTxid == ""`;
- `recoverBtcToArkadeChainSwap` at recover.go:265.

So the sequence "user is slow, funds the lockup address after `defaultChainSwapTimeout` (30m)" leaves BTC locked in the HTLC with no automated recovery, and `RefundChainSwap` refuses to run with `btc lockup of swap %s was never funded, nothing to refund` precisely when it is needed. The ephemeral refund key is in the store, so the funds are not lost, but the SDK gives no way to reach them.

30 minutes is not a lot of margin for a human doing an onchain send, so I do not think this is a corner case.

### Suggested fix

Do not trust the stored txid as the source of truth for "was it funded". Fall back to asking the explorer, since the HTLC address is already persisted in `swap.ChainSwap.Address`:

```go
if swap.ChainSwap.FundingTxid == "" {
    // The lockup may have landed after the watcher goroutine exited, in which case nothing
    // recorded its txid. Ask the explorer before giving up.
    htlcScript, refundKey, err := h.getSwapHtlc(ctx, swap)
    if err != nil {
        return nil, nil, err
    }
    txid, err := h.findBtcLockup(ctx, swap, htlcScript)
    if err != nil {
        return nil, nil, fmt.Errorf(
            "btc lockup of swap %s was never funded, nothing to refund: %w", swapId, err,
        )
    }
    swap.ChainSwap.FundingTxid = txid
    if err := h.persistUpdatedSwap(ctx, *swap); err != nil {
        return nil, nil, err
    }
}
```

where `findBtcLockup` looks up UTXOs at `swap.ChainSwap.Address` via `h.wallet.Explorer()` and picks the one paying `htlcScript.PkScript(false)`. `refundBtcLockup` already does the matching-output scan once it has a tx, so this only needs to supply the txid.

The same fallback applied in `recoverPendingSwaps` would make it recover on restart without the caller having to notice. Alternatively, or additionally, keep watching the lockup address for a while past the swap timeout rather than tying the watcher's lifetime to it.

bitcoin-coder-bob:
Lets add an e2e test that drives `ArkadeToBtcChainSwap` end to end, kills the manager mid flight, and then do a recover

altafan:
> Lets add an e2e test that drives `ArkadeToBtcChainSwap` end to end, kills the manager mid flight, and then do a recover

This is exactly waht we can't support. We support recover, ie finalizing a chain swap with this direction if it stopped in the middle of the exuction, and we already have the e2e test for it. We can restore such chain swap, but we couldn't finalize it becuase the key to claim/refund the onchain htlc was ephemeral and it's now lost forever

bitcoin-coder-bob:
> it becuase the key to claim/refund the onchain htlc was ephemeral and it's now lost forever

Yeah agreed on restore but that iss not what Im going for here. A restored ark -> btc swap can only ever be refunded. The btc claim key is ephemeral and boltz never sees it. restore.go:172 already says exactly that in your comment on that line.

What I mean is the recover path which is a different thing. `restoreSwaps` bails out unless the store is empty (restore.go:36 "Restore only makes sense on an empty store"). `recoverPendingSwaps` runs off the local swap.db and the ephemeral key is still sitting right there we persist it at chainswap.go:180:

```
  PrivateKey: hex.EncodeToString(claimKey.Serialize()),
```

 `recoverChainSwapClaim` decodes it and rebuilds the htlc. So stop the manager then start a fresh one on the same datadir and nothing is lost. That is the test I mean to be done.

on the existing coverage `TestRecoverPendingSwaps/claims` a chain swap left pending on shutdown never actually calls ArkadeToBtcChainSwap. It goes through `createBoltzChainSwap`, which writes the record including `Preimage: preimage`. So swapPreimage returns on its first line and the fallback under it never runs.

That is why the `persistPreimage := !ok` regression I flagged doesnt show up in CI. With an HD identity the preimage isnt persisted anymore, and the contract has SenderKeyId set so ReceiverKeyId is empty and the fallback bails. The blocker there is the preimage and not the key.

altafan:
@bitcoin-coder-bob Fixed in both directions since your comment: `swapPreimage` now re-derives from the sender key for Arkade→BTC (persisting stays single-key-identity-only), and the e2e helper no longer seeds the preimage — it derives it the same way the manager does — so the existing recovery test exercises the exact fallback that was broken. The staged helper intentionally reproduces the mid-flight crash state deterministically; it may be tricky to find the right time to kill the manager in the middle of a `ArkadeToBtcChainSwap` call

bitcoin-coder-bob (approved):

PR #217, opened 2026-08-04
swap: claim the vtxo boltz's lockup tx created

Kukks:
Both internal claim paths passed a nil outpoint to `selectClaimableVTXO`, so selection fell back to creation order. Ordering picks by age, which means anything that reached the VHTLC address before boltz funded it gets claimed first.

Boltz names the locking transaction in the update that triggers the claim, and both paths already have it — the chain swap handler reads it for `ServerLock`, the reverse swap loop ignores it. Resolve that txid to the vtxo it created and claim that one.

`outpointForFundingTx` returns:

| input | result |
|---|---|
| empty txid | nil outpoint, selection unchanged |
| txid matches one vtxo | that outpoint |
| txid matches nothing yet | `ErrorNoVtxosFound` |
| txid paid the address twice | error, nothing distinguishes the outputs |

The third row is the one to look at. Returning `ErrorNoVtxosFound` rather than a nil outpoint means the reverse swap keeps waiting for the funding vtxo instead of falling back to whatever is already at the address — which is why resolution happens *inside* the retry closure rather than before it. The chain swap handler has no retry and already fails when no vtxo is found, so its behaviour there is unchanged.

Also drops the `//TODO check if we should pass outpoint similarly as in swap` in the chain swap handler. The answer was yes, and the swap path wasn't doing it either.

## Test plan

`selectOutpointByFundingTxid` is split out from the indexer lookup so the selection can be tested directly. Eight cases: empty txid, no match, no vtxos at all, match in spendable, match in pending, the same outpoint reported in both lists, one tx paying the address twice, and ambiguity split across the two lists.

The both-lists case is the subtle one — a vtxo reported as both spendable and pending is one candidate, not competing outputs.

- `./swap/...`: ok, matching the pre-change baseline
- `go build ./...`, `go vet ./swap/...` clean
- gofmt clean

## Note

The same change landed in fulmine as ArkLabsHQ/fulmine#466, where the integration suite covers both paths end to end (`TestReverseSwap`, `TestChainSwapBTCtoARK`) and passes. This repo has no equivalent integration coverage for those flows, so the unit tests above are the gate here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap claims by selecting the precise funding output associated with each transaction.
  * Added retries when the matching output is temporarily unavailable.
  * Prevented claims from using unrelated or ambiguous funds.
  * Swaps now fail clearly when the required funding output cannot be identified.
  * Improved Bitcoin Ark swap claims by resolving the associated funding output before claiming.

### DIFF
--- a/swap/chainswap_btc_ark_handler.go
+++ b/swap/chainswap_btc_ark_handler.go
@@ -122,13 +122,22 @@ func (b *btcToArkHandler) handleBtcToArkServerLocked(
 	serverLockupTxID := update.Transaction.Id
 	b.chainSwapState.Swap.ServerLock(serverLockupTxID)
 
+	// Claim the vtxo that boltz's lockup tx created, rather than whichever one
+	// sorts first by age at this address.
+	outpoint, err := b.swapHandler.outpointForFundingTx(
+		ctx, b.chainSwapState.Swap.VhtlcOpts, serverLockupTxID,
+	)
+	if err != nil {
+		b.chainSwapState.Swap.Fail(fmt.Sprintf("claim failed: %v", err))
+		return fmt.Errorf("failed to resolve Ark VTXO for lockup tx %s: %w", serverLockupTxID, err)
+	}
+
 	// Claim Ark VTXOs lockup
-	//TODO check if we should pass outpoint similarly as in swap
 	claimTxid, err := b.swapHandler.ClaimVHTLC(
 		ctx,
 		b.preimage,
 		b.chainSwapState.Swap.VhtlcOpts,
-		nil,
+		outpoint,
 	)
 	if err != nil {
 		// ChainSwap.Fail() emits FailEvent automatically
--- a/swap/chainswap_btc_ark_handler.go
+++ b/swap/chainswap_btc_ark_handler.go
@@ -124,6 +124,14 @@ func (b *btcToArkHandler) handleBtcToArkServerLocked(
 
 	// Claim the vtxo that boltz's lockup tx created, rather than whichever one
 	// sorts first by age at this address.
+	//
+	// This fires on the mempool event, so the vtxo may not be indexed yet and
+	// resolution can report ErrorNoVtxosFound. There is no retry here and the
+	// swap fails, which is the behaviour that was already in place: claiming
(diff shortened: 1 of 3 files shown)


Kukks:
Two addressed in `a548bd2`, three answered below.

**[4] `ErrorNoVtxosFound` sentinel construction.** Fixed. `errors.New` now, with a comment saying it is a sentinel compared by `errors.Is`. You are right that it worked only by identity; the risk was someone later writing `fmt.Errorf("...: %v", ErrorNoVtxosFound)` and silently turning a retry into a failure.

**[2] Chain swap has no retry on resolution.** Comment added rather than a retry. Your reading is right that behaviour is unchanged — claiming with a nil outpoint reached the same error from `selectClaimableVTXO` when nothing was at the address — so adding a retry here would be a new behaviour introduced under cover of a fix, on a timing window neither of us has evidence for. The comment now records that, and what the fix would be if the window proves real. If the integration suite shows it, I would rather do it as its own change with a test that reproduces the timing.

**[1] Double indexer round-trips.** Real, and the arithmetic is right: 4 calls per 200ms tick where 2 would do, in exactly the window where the vtxo is not yet visible. Not fixing it here. Collapsing it means either threading the resolved outpoint into `selectClaimableVTXO` or giving `ClaimVHTLC` a selector, and both change signatures shared with the refund and settle paths. Folding a signature change across three flows into a fix whose point is selection correctness is how a small fix becomes hard to review. Worth doing as a follow-up.

**[3] TOCTOU error sentinel asymmetry.** Confirmed and deliberately not changed. If the vtxo disappears between resolution and claim, `selectClaimableVTXO` returns a plain formatted error rather than `ErrorNoVtxosFound`, so `waitAndClaim` treats it as fatal instead of retrying.

Making it wrap the sentinel would fix that case and break a more common one: a caller passing a genuinely wrong outpoint would then retry until the context deadline instead of failing immediately with a usable message. Given the case is, as you say, practically impossible — we hold the preimage, nobody else can legitimately claim the vtxo — trading a fast, clear failure for a silent timeout is the worse side of that trade. Flagging it rather than changing it.

**[5]** Nothing to do, thanks for checking the embedded-struct assumption in `vtxoAt` — that was the part of the helper I was least sure of.

